PR #86, opened 2021-09-02
Even BETTER and more flexible screenshake

grufkork:
Based off of previous rework, but also keeping the rotational shake. Message from last time:

"Reworked screenshake so it looks better and is more flexible. I replaced the rotational shake with translational, which is how it's usually done. There's two shake modes, sinusodial and noise, of which only noise is used at the moment. Sinusodial has a springier feeling, and could work for hard landings or such.
[...]

For the noise, I ported a javascript library I have used on a couple other of projects. I have got no idea how it performs, but it works just fine for this. Looking at the output though it seems to sometimes generate short, flat plateaus, but it might also just be that noise looks that way. [...]"


not-fl3:
Thanks! :+1:

PR #92, opened 2021-09-06
New explosion VFX and standardised function

grufkork:
The explode-function does not yet actually blow anything up, but the particles are pretty


src/main.rs
```diff
@@ -128,12 +132,19 @@ impl Resources {
...
-
+        const CANNONBALL_HIT_FX: &str = include_str!("../assets/fxses/canonball_hit.json");
+        const EXPLOSION_PARTICLES: &str = include_str!("../assets/fxses/explosion_particles.json");
+        const SMOKE_FX: &str = include_str!("../assets/fxses/smoke.json");
         let hit_fxses = EmittersCache::new(nanoserde::DeJson::deserialize_json(HIT_FX).unwrap());
         let explosion_fxses =
             EmittersCache::new(nanoserde::DeJson::deserialize_json(EXPLOSION_FX).unwrap());
         let life_ui_explosion_fxses =
             EmittersCache::new(nanoserde::DeJson::deserialize_json(LIFE_UI_FX).unwrap());
+        let fx_cannonball_hit =
```
not-fl3, line 143: Will cannonball fx be used in too many weapons to be loaded here, in global resources list? 

(it may be, but from the first look it belongs to weapon's fx list more)

src/nodes/bullets.rs
```diff
@@ -60,7 +60,6 @@ impl scene::Node for Bullets {
     }
```
not-fl3, line 60: Ohhh, I forgot to delete bullets.rs :( 
But it was not compiling anyway - it was not it nodes.rs

src/utils.rs
```diff
@@ -0,0 +1,34 @@
...
+    //Center position, radius of explosion. Only FX at the moment
+    let mut r = 0.0;
+
+    let mut resources = storage::get_mut::<Resources>();
+    while r < radius - 5. {
+        // Particles are 5 in radius
+        r += 10.0 / (r + 1.);
+        let angle = gen_range(0.0, f32::consts::PI * 2.);
+        resources
+            .fx_cannonball_hit
```
not-fl3, line 21: name `fx_cannonball` assumes that the fx is used only by the cannon weapon
while `explode` function looks like it is designed for use in any weapon

Maybe `explode` should be moved to a cannon weapon?

grufkork: I've just kept the name out of convention, since that's what it was named previously :) Nothing else uses it now so renaming it should be no issue

grufkork:
Closing due to #113

Issue #72, opened 2021-08-05
New keyboard mappings


64kramsystem:
With the new functionality of the platform descension, the old mappings are not practical anymore (ie. pressing up and down arrow at the same time, or W+S); the key `K` is also temporarily mapped to the second player (formerly, it was mapped to the first player).

Given the two above, it's time for a refresh, to find new, sensible, defaults.

zicklag:
Closing this as I think it's been resolved.

PR #93, opened 2021-09-07
improve default key bindings

joseluis:
This is my proposal for improved default key bindings for both player 1 and player 0.

related: #72


not-fl3:
Thanks!

PR #94, opened 2021-09-09
Re-add the sounds, apply where last left off

ArcOfDream:
Take Twoie. Seems some weapons aren't here in the rewrite yet, so only the player stuff's been implemented.


src/main.rs
```diff
@@ -97,6 +110,19 @@ impl Resources {
         let shoot_sound = load_sound("assets/sounds/shoot.ogg").await?;
         let sword_sound = load_sound("assets/sounds/sword.wav").await?;
         let pickup_sound = load_sound("assets/sounds/pickup.wav").await?;
+        let swing_sound = load_sound("assets/sounds/swing.wav").await?;
```
not-fl3, line 113: Hi! 
Now items use their own resources, not the resources from `main`, like here: https://github.com/fishfight/FishFight/blob/main/src/items.rs#L41

So weapon's sounds should be in the  per-item `sounds` list rather than here
(but let me double check, maybe I forgot to actually load the sounds from this list)

not-fl3: Oh, I just noticed that this PR is only about two sounds - player landing and player throw. 
Yes, it is totally fine to add them here to resources. 
But could you remove the other, unrelated to the PR, sounds?

ArcOfDream: Also player death. Currently haven't implemented any item sounds, but that's good to know.
Will remove item related sounds from the resources struct. Should I move the original weapon sounds (shoot, sword) to their respective items, too?

ArcOfDream: Also by removing, do you want me to also take out the sounds from the assets folder, or should I leave them be? :thinking:

not-fl3: > Also by removing, do you want me to also take out the sounds from the assets folder, or should I leave them be? thinking

Up to you, I think it is convenient to store them in git
but if you have some other storage - feel free to add only the used ones

ArcOfDream: Cool, I think I'll keep the stuff in the folder.

.gitignore
```diff
@@ -2,3 +2,4 @@
 src/.DS_Store
 .DS_Store
 /.idea/
+/.vscode/
```
not-fl3, line 5: Some of the local settings somehow sneaked into our .gitignore, but lets not make it a rule :) 

I would prefer to keep `*~`, `*.bak`, `thumbs.db` or other similar files like editor settings in a global .gitignore, https://stackoverflow.com/questions/7335420/global-git-ignore

ArcOfDream: Whoops. Fixed that.

not-fl3: 👍

not-fl3:
LGTM!
Just make the CI happy again :)

PR #104, opened 2021-09-20
Cannon reimplementation

grufkork:
Added back the cannon. Cannonball texture is placeholder. Also contains the changes from the standardised-explosions branch. Or so it seems. I and git are not friends. 

Cannon is not networked yet, but it might quite possibly work out of the box.


src/circle.rs
```diff
@@ -0,0 +1,59 @@
+use macroquad::math::{vec2, Rect, Vec2};
```
not-fl3, line 1: This code got moved to https://docs.rs/macroquad/0.3.10/macroquad/math/struct.Circle.html#
If macroquad's circle is sufficient - I would prefer using it

assets/fxses/canonball_hit.json
```diff
@@ -1 +1,51 @@
-{"local_coords":false,"emission_shape":"Point","one_shot":true,"lifetime":0.2,"lifetime_randomness":0,"explosiveness":0.65,"amount":41,"shape":{"Circle":{"subdivisions":10}},"emitting":false,"initial_direction":{"x":0,"y":-1},"initial_direction_spread":6.2831855,"initial_velocity":73.9,"initial_velocity_randomness":0.2,"linear_accel":0,"size":64.0,"size_randomness":0.0,"blend_mode":"Alpha","colors_curve":{"start":{"r":0.8200004,"g":1,"b":0.31818175,"a":1},"mid":{"r":0.71000004,"g":0.36210018,"b":0,"a":1},"end":{"r":0.02,"g":0,"b":0.000000007152557,"a":1}},"gravity":{"x":0,"y":0},"post_processing":{}}
+{
```
not-fl3, line 1: I am not so sure of pretty-printed jsons here. Usually they are copy-pasted to/from the editor back and forth all the time, and pretty-printing on each commit would be some extra work?

src/nodes/cannonball.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        let mut body = PhysicsBody::new(
+            &mut resources.collision_world,
+            pos,
+            0.0,
+            vec2(Self::CANNONBALL_WIDTH, Self::CANNONBALL_HEIGHT),
+        );
+        body.speed = speed;
+
+        /*PhysicsBody {
```
not-fl3, line 85: Git is pretty OK for keeping older version of the code, I would prefer to avoid commented for historical purposes code

src/nodes/cannonball.rs
```diff
@@ -0,0 +1,204 @@
...
+        }
+    }
+
+    pub fn spawn(pos: Vec2, facing: bool, owner_id: u8) {
+        let cannonball = Cannonball::new(pos, facing, owner_id);
+        scene::add_node(cannonball);
+    }
+}
+
+/*impl EruptedItem for Cannonball {
```
not-fl3, line 127: Same situation here

If the code is temporary turned off for the sake of PR - it may be worth doing an additional draft PR with said code, but keeping the PR itself clean.

In some cases it really is easier to just comment out something, but, at least, leave a comment on why the code is commented (and I feel that right now this is not the case, and commented code just do not belong to the feature implemented by this PR)

src/nodes/cannonball.rs
```diff
@@ -0,0 +1,204 @@
...
+        &mut self.body
+    }
+    fn enable_at_y(&self) -> f32 {
+        self.erupting_enable_on_y.unwrap()
+    }
+}*/
+
+impl scene::Node for Cannonball {
+    fn fixed_update(mut cannonball: RefMut<Self>) {
+        /*if cannonball.erupting {
```
not-fl3, line 150: Same, commented code :)

src/utils.rs
```diff
@@ -0,0 +1,54 @@
+use macroquad::{
```
not-fl3, line 1: while this is used just by one weapon - I would consider moving it to this weapon and generalize later

assets/fxses/canonball_hit.json
grufkork: Oh, I found the tool generates settings in a slightly different format, such as blend_mode being an object instead of just a string. I formatted it to check for differences to the others, but I'll fix it!

src/circle.rs
grufkork: Ah right, must've slipped through when I merged

src/nodes/cannonball.rs
grufkork: It's just there for when the volcano is reimplemented, but that will probably need some new code anyways. Removing

src/utils.rs
grufkork: It's about to be used for the mines (being worked on by @kindaro (it's here now!)), so is it fine if I leave it as is? I'll also probably use it for the kickbombs or something like that next.

src/nodes/cannonball.rs
```diff
@@ -0,0 +1,140 @@
+use macroquad::{
```
not-fl3, line 1: Is cannonball going to be used for many weapons? I feel that it belong to a cannon item, not to the whole game?

erlend-sh:
Gonna repost @grufkork’s Discord message here:

> Righto, I had a free hour this afternoon so I've converted the cannonball to a generic explosive, though I just got two questions before pushing:
1. To not make the spawn function require five metres of parameters I moved the detonation settings to a struct within the new explosives.rs file. However, nodes.rs only imports the Explosive struct. I solved this by bunching DetonationParameters and Explosive into a mod, and then import the whole thing. This however requires that every field in the DetonationParameters struct is marked public, which seems ugly. I guess there's a rustier way to do this?
> 
> 2. Mostly a question of preference and convention. For some settings it's either a value or not at all (no proximity fuze vs prox fuze radius, no fuse vs. fuse length). I see three solutions to this: 
a. have a magic value (eg. 0) to indicate not using the feature and check with an if for that value. Conserves bits but isn't clear what it does without reading documentation
b. have a separate boolean for whether it's enabled or not. Means parameters must be set that are not used. ((((also uses a byte))))
c. Use an enum. Seems excessively wordy though. (Also uses some extra memory?)

@64kramsystem might have some input if he’s around.

src/utils.rs
not-fl3: I would prefer to copy-paste this for each weapon (1 or 2 its not a big deal), but up to you, I can live with utils.rs

src/nodes/cannonball.rs
erlend-sh: It could also appear individually from something like a “weapons volcano” and other spawn behaviors.

src/utils.rs
grufkork: If the explosive class is used for all explosive weapons, the util class should no longer be necessary. I'll move it

grufkork:
Closing due to #113

PR #105, opened 2021-09-20
Turtle Shell reimplementation

Chosenundead15:
reimplemantation of the old code functionality, adapted to use the scene::add_node and spawn functions. The shell stops two shots from the back as it previously did.

### DIFF
--- a/src/items.rs
+++ b/src/items.rs
@@ -5,7 +5,7 @@ mod machine_gun;
 mod muscet;
 mod sproinger;
 mod sword;
-
+mod turtle_shell;
 use macroquad::{experimental::scene::HandleUntyped, math::Vec2};
 
 /// Proto-mod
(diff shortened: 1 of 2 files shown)


src/items.rs
```diff
@@ -5,7 +5,7 @@ mod machine_gun;
 mod muscet;
 mod sproinger;
 mod sword;
-
+mod turtle_shell;
```
not-fl3, line 8: Looks like turtle_shell.rs file is missing :(

Chosenundead15: that was odd, readded the files to the commit, it is saying there is a clean tree so there shouldn't be an issue now

not-fl3: :+1: 
No worries, this happens quite often. I usually check "Files changed" tab here on GH while opening PR to double check that I am commiting the files I really want to be changed by the PR :)

not-fl3:
Thanks! :+1:

PR #112, opened 2021-09-23
Shoes reimplementation

legendiguess:
In compare to old shoes implementation i added `PhysicsBody` and `ThrowableItem` to it, and made that shoes drops from the killed fish.

Closes https://github.com/fishfight/FishFight/issues/108

### DIFF
--- a/src/items.rs
+++ b/src/items.rs
@@ -3,6 +3,7 @@
 
 mod machine_gun;
 mod muscet;
+pub mod shoes;
 mod sproinger;
 mod sword;
 mod turtle_shell;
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -87,6 +88,9 @@ impl Resources {
         let turtleshell = load_texture("assets/Whale/TurtleShell(32x32).png").await?;
         turtleshell.set_filter(FilterMode::Nearest);
 
+        let shoes = load_texture("assets/Whale/Shoes(32x32).png").await?;
```
not-fl3, line 91: Isnt item's texture enough?

not-fl3: For reference, check https://github.com/fishfight/FishFight/blob/main/src/items/sword.rs#L50

PR #113, opened 2021-09-23
Standardised explosives

grufkork:
Draft for a generic explosive node. Used by cannonballs, should be easily repurposed for grenades and mines.


src/utils.rs
```diff
@@ -0,0 +1,54 @@
+use macroquad::{
+    experimental::collections::storage, experimental::scene, experimental::scene::RefMut,
+    prelude::Vec2, rand::gen_range,
+};
+
+use crate::{nodes::Player, Circle, Resources};
+use std::f32;
+
+/// Creates explosion FX and kills players within `radius`
+pub fn explode(position: Vec2, radius: f32) {
```
not-fl3, line 10: "explode" is still used in exactly one place. I still think that it belong to, now, "explosive.rs"

src/nodes/explosive.rs
```diff
@@ -0,0 +1,144 @@
+pub mod explosives {
```
not-fl3, line 1: Why its a mod within a mod? Maybe just remove this line?

not-fl3:
Is this intended to be merged on top or instead https://github.com/fishfight/FishFight/pull/104 ?
If on top - maybe rebase this PR on top of `cannon-reimplementation-fixed` branch?

grufkork:
Well, all the previous pulls are based of each other. I'll close those

src/nodes/fxses.rs
```diff
@@ -4,6 +4,7 @@ use macroquad::{
         collections::storage,
         scene::{self, RefMut},
     },
+    prelude::Vec2,
```
not-fl3, line 7: maybe `math::Vec2`?

src/nodes/explosive.rs
grufkork: That's for accessing the `detonation_parameters` struct from elsewhere… Otherwise I would have to add both to nodes.rs, which doesn't seem right. Though this doesn't really seem like the right solution either, especially since every field needs to be marked `pub`.

not-fl3: but it will be equally available without a mod within mod?

grufkork: The mod is for grouping, so there aren't any non-node structs in nodes.rs. But it was mostly a temporary solution. Have you got some cleaner solution? Of course, I could just add it straight to nodes.rs but since it is closely tied to the explosives they ought to stick together somehow. Though the mod makes the explosives unnecessarily isolated...

not-fl3: In nodes.rs there are only things you explicitly decided to `use`. Therefore if you do not use something in `nodes.rs` - it will not make it there.
If you want a module, not struct being available in `nodes` - do not use anything, do `pub mod explosive`.

kindaro:
I tested this and it appears to me that setting fuse to 0 results in immediate explosion rather than the mine staying there forever. Possibly this field should have type `Option` for clarity? `Some 0.` would indicate immediate explosion, while `None` would indicate that the explosive only explodes when triggered.

grufkork:
I realise I never actually tested that... Options are probably the cleanest way to handle it. I'm fairly busy the next days but I might be able to sneak in some work tonight and fix it.

grufkork:
There we go. This should be done now, unfused explosives work correctly. Also I figured out how imports and stuff works so explosive.rs doesn't use a mod any more.

src/nodes/explosive.rs
grufkork: It seems I overcomplicated things, got it figured out now!

src/utils.rs
grufkork: Moved

kindaro (approved):
Works for me!

* With `fuse: Some (0.)`, the explosive explodes immediately.
* With `fuse: None`, the explosive stays around indefinitely, waiting to be triggered.

This aligns with my expectations.

* * *

I did not look at the rest of the code thoroughly — I do not really know Rust anyway.

With a pedant hat on, I should suggest splitting the pull request into 2:

* Add standardized explosives.
* Add the cannon.

I should also suggest squash merging.

grufkork:
Alright, the commits contain changes for both cannon and explosive so I'll just squash it all. But I'll keep it in mind for next time!

PR #114, opened 2021-09-23
Grenades

Chosenundead15:
the grenades are functioning without issues.


src/components/erupted_item.rs
```diff
@@ -0,0 +1,49 @@
...
+use macroquad::prelude::{collections::storage, get_frame_time};
+
+use crate::Resources;
+
+use crate::components::PhysicsBody;
+
+//at the moment is not possible to test the code as ite requires EruptingVolcano to test
+//It also is dependency of ArmedGrenade so a rough estimation of the code to not implement the trait later
+//is here.
+pub trait EruptedItem {
```
not-fl3, line 11: If the trait is not used anywhere - maybe exclude it from this PR? And add as a separate PR or withing the PR that is going to use it?

Chosenundead15: Is "used" for the grenades, but the code can't be tested because it requires a functioning volcano. So for the separate PR, I think it should eliminated the trait from grenades.rs and put the version which has it implemented on the separate PR as you said.

not-fl3: This PR is about Grenade item and EruptItems (especially untested) looks way out of scope of this PR as for

Chosenundead15: done, moved away from the PR

src/components/armed_grenade.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    pub fn spawn(pos: Vec2, facing: bool) {
+        let grenade = ArmedGrenade::new(pos, facing);
+        scene::add_node(grenade);
+    }
+}
+
+impl Node for ArmedGrenade {
```
not-fl3, line 84: The idea of `components` - its a piece of logic that may help building nodes (described here: https://github.com/fishfight/FishFight/blob/main/src/components.rs#L1)

As for me, your `ArmedGrenade` struct belong more to an Item?

Chosenundead15: good point, at the time I was associating it with bullets which was a component to make firearms. ArmedGrenade is prety much the projectile of grenades but also there is not a common explosive component so it makes it a not a proper component, but similar to an item at the moment.

not-fl3: Bullets lives in their items, like muscet's bullet lives in muscet.rs, so maybe put ArmedGrenade near Grenade item as well?

Chosenundead15: I don't think I follow that.

not-fl3: ArmedGrenade is not an item by itself, it is an implementation detail of a Grenade, an item. 

Therefore it belongs to a /grenade/armed_grenade.rs
or just in grenade.rs

Just as muscet's bullet lives in muscet.rs

Chosenundead15: Oh, I get what you mean now, yeah that makes sense and should work better that way

not-fl3:
:tada: 
Thanks!

PR #117, opened 2021-09-25
Put mines back.

### LINKED ISSUE #99: Put mines back.
Mines were added in #8. They were later removed, but I am going to put them back, as fatal and insidious as ever.

* [ ] Adjustable trigger delay.
* [x]  Trigger radius independent of explosion radius.
  * [ ] Trigger radius larger than explosion radius?
* [ ] Explosions trigger mines in their radius?

kindaro:
Resolve #99.

* Checked out from 3ba21b6185114a5c69c791c12fc1e0884bfcb102 — the last
  commit where this weapon existed.

* Uses work in progress from #113 for explosive projectiles. Since that work is behind `main` by a few commits, I simply copied the files I needed. This patch will become half its current size once #113 is merged.

### DIFF
--- a/src/items.rs
+++ b/src/items.rs
@@ -4,6 +4,7 @@
 mod cannon;
 mod grenades;
 mod machine_gun;
+mod mines;
 mod muscet;
 pub mod shoes;
 mod sproinger;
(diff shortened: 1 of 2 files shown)


kindaro:
I chose to qualitatively deviate from the original behaviour in one instance: in this draft mines are thrown with velocity 0. I think this is more cool because it communicates that they are attached to the floor and makes them dissimilar from grenades. Please comment on this!

kindaro:
Let us wait for #113 to get merged.

grufkork:
Added the new explosive system so it doesn’t compile now, but that should be easily fixed!

kindaro:
Ready to be merged!

PR #125, opened 2021-09-27
Replace quad-gamepad with gilrs

### LINKED ISSUE #122: Pressing media keys on keyboard crashes the game on Linux
Pressing any of the media keys (volume up/down, play/pause, etc.) on my keyboard crashes the game. The media keys seem to be mapped to their own evdev device, and Fish Fight detects that device as a gamepad.

Environment: Arch Linux 5.14.7, Xfce with Xorg

### LINKED ISSUE #123: Poor responsiveness with some gamepads
The game does not work well when using the gamepads listed below. Inputs are frequently missed, stuck, or delayed.

- Sony DualSense
- Microsoft Xbox One pad

PotatoTech:
Resolves #122, resolves  #123.

This still needs to be tested on multiplayer. Additionally, there are a few new issues caused by this change that may need to be fixed:
- Action pad buttons use north/south/east/west internally instead of ABXY. This forces controllers using the Nintendo layout to use the Xbox layout instead.
- ~Controllers that do not have analog sticks can no longer control the player using the D-pad. This can be fixed by manually mapping the D-pad to the same controls as the left analog stick.~ (Fixed)


PotatoTech:
`libudev.pc` is a build dependency for GilRs on Linux. It is provided by the `libudev-dev` package on Ubuntu but is currently absent in the CI environment. If that could be added somehow, the checks should pass.

erlend-sh:
Could you compare build times for our status quo vs the introduction of gilrs? We can take on gilrs, but we should make a clear note of the costs that come with it.

It’s also worth noting that gilrs is no longer actively maintained, so if we need to fix something there we need to do it ourselves, and we may need to maintain a fork.

PotatoTech:
Based on the times reported by the Github workflow, build times are longer with GilRs, especially on macOS.

| Platform | Status Quo Time | GilRs Time |
| --- | --- | --- |
| ubuntu-latest, x86_64-unknown-linux-gnu | 23s | 20s
| ubuntu-latest, x86_64-unknown-linux-gnu, --release | 36s | 52s
| macos-latest, x86_64-apple-darwin | 17s | 30s
| macos-latest, x86_64-apple-darwin, --release | 54s | 2m 14s
| windows-latest, x86_64-pc-windows-msvc | 46s | 45s
| windows-latest, x86_64-pc-windows-msvc, --release | 1m 13s | 1m 23s

erlend-sh:
The comparison is indeed very helpful, thanks! 

Some pretty significant increases here. But we’ve logged them now, so we’re fine to proceed to see if the compatibility gains are worth it. 👍

PotatoTech:
It seems that the workflow build times on my fork were not accurate. Here are the updated GilRs timings taken from the checks run from here:

| Platform | Status Quo Time | GilRs Time |
| --- | --- | --- |
| ubuntu-latest, x86_64-unknown-linux-gnu | 23s | 26s
| ubuntu-latest, x86_64-unknown-linux-gnu, --release | 36s | 43s
| macos-latest, x86_64-apple-darwin | 17s | 16s
| macos-latest, x86_64-apple-darwin, --release | 54s | 58s
| windows-latest, x86_64-pc-windows-msvc | 46s | 52s
| windows-latest, x86_64-pc-windows-msvc, --release | 1m 13s | 1m 24s

olefasting:
As I mentioned on the issue, I have problems with mappings with gilrs in my framework. I have some very strange mappings on my gamepad, which seem to me to be so out of place, that I can only conclude that they vary from gamepad to gamepad. If it is the case that gilrs has problems in terms of mappings varying greatly from gamepad type to gamepad type, I believe should not merge this, until we have a way of mapping keys, either via GUI or, at the very least, via a config file

PotatoTech:
You seem to be right. I tested with a USB SNES controller, and the D-pad does not work correctly (only the right button works). Perhaps I will have better luck with using the SDL2 bindings directly. Closing this for now.

PR #131, opened 2021-09-28
Reimplement Galleon Weapon

Zac8668:


src/items/galleon.rs
```diff
@@ -0,0 +1,317 @@
...
+        );
+
+        (pos, facing)
+    }
+
+    pub fn update(&mut self) -> bool {
+        self.pos += self.speed * get_frame_time();
+        self.lived += get_frame_time();
+
+        if self.lived > Self::LIFETIME {
```
64kramsystem, line 80: The lifetime logic can be entirely removed, by performing the check based on the position.

Pseudocode:

```
if object.x >= SCREEN_WIDTH || object.x <= - OBJECT_WIDTH {
  # object is dead
}
```

Zac8668: I removed it, but I used the map width as base to removing the flying galleon.

src/items/galleon.rs
```diff
@@ -0,0 +1,318 @@
...
+                && !player.dead
+            {
+                scene::find_node_by_type::<crate::nodes::Camera>()
+                    .unwrap()
+                    .shake_noise(1.0, 10, 1.);
+                {
+                    let mut resources = storage::get_mut::<Resources>();
+                    resources.hit_fxses.spawn(player.body.pos);
+                }
+                let direction = self.pos.x > (player.body.pos.x + 10.);
```
64kramsystem, line 101: Converting hardcoded values to constants makes the code considerably more readable, IMO.

When I approached the code for the first time, all the hardcoded values were... "not helpful" :grin:

src/items/galleon.rs
```diff
@@ -0,0 +1,318 @@
...
+
+        start_coroutine(coroutine)
+    }
+
+    pub fn spawn(pos: Vec2) -> HandleUntyped {
+        let mut resources = storage::get_mut::<Resources>();
+
+        let sprite = GunlikeAnimation::new(
+            AnimatedSprite::new(
+                32,
```
64kramsystem, line 170: Constants should be used here. There are a few options, e.g. rename `COLLIDER_*` to something like `SPRITE_*`, or just using `COLLIDER_*` - your choice :slightly_smiling_face:

src/items/galleon.rs
```diff
@@ -0,0 +1,318 @@
...
+            let mut node = scene::get_untyped_node(node).unwrap().to_typed::<Galleon>();
+
+            node.body.angle = 0.;
+            node.throwable.owner = Some(owner);
+        }
+
+        fn mount(node: HandleUntyped, parent_pos: Vec2, parent_facing: bool) {
+            let mut node = scene::get_untyped_node(node).unwrap().to_typed::<Galleon>();
+            let mount_pos = if node.body.facing {
+                vec2(15., 16.)
```
64kramsystem, line 230: Other good candidates for constants.

src/items/galleon.rs
```diff
@@ -0,0 +1,318 @@
...
+                self.pos.y,
+                Self::WIDTH,
+                Self::HEIGHT,
+            )) && player.id != self.owner_id
+                && !player.dead
+            {
+                scene::find_node_by_type::<crate::nodes::Camera>()
+                    .unwrap()
+                    .shake_noise(1.0, 10, 1.);
+                {
```
64kramsystem, line 97: I think braces are redundant here, unless there is some specific reason.

src/items/galleon.rs
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    pub fn update(&mut self) -> bool {
+        self.pos += self.speed * get_frame_time();
+
+        {
+            let resources = storage::get::<Resources>();
+            let map_width = (resources.tiled_map.raw_tiled_map.tilewidth
+                * resources.tiled_map.raw_tiled_map.width) as f32;
+            if self.pos.x + Self::WIDTH < -map_width * 0.5 || self.pos.x > map_width * 1.5 {
```
64kramsystem, line 80: What's the reason for not using the following condition?

```rs
self.pos.x + Self::WIDTH <= 0 || self.pos.x >= map_width
```

if there is a specific reason, I think it'd be good to add it to a comment, since it's not obvious when reading the code.

64kramsystem (approved):
Code looks good. My notes are just about good practices. Approvals for merge, and merge, are done by Fedor :slightly_smiling_face:

src/items/galleon.rs
Zac8668: Actually there is, I can't get resources without an error if I don't use them.

src/items/galleon.rs
Zac8668: When both players are on the far left or far right of the map, the outside is visible, if I changed to the code above then it would be visible when the Flying Galleon vanished, that's why I put it to vanish after it crossed half the map width off bounds.

src/items/galleon.rs
64kramsystem: This diff applies without errors:

```diff
diff --git i/src/items/galleon.rs w/src/items/galleon.rs
index dc21f74..e5ad0e3 100644
--- i/src/items/galleon.rs
+++ w/src/items/galleon.rs
@@ -94,10 +94,10 @@ pub fn update(&mut self) -> bool {
                 scene::find_node_by_type::<crate::nodes::Camera>()
                     .unwrap()
                     .shake_noise(1.0, 10, 1.);
-                {
-                    let mut resources = storage::get_mut::<Resources>();
-                    resources.hit_fxses.spawn(player.body.pos);
-                }
+
+                let mut resources = storage::get_mut::<Resources>();
+                resources.hit_fxses.spawn(player.body.pos);
+
                 let direction = self.pos.x > (player.body.pos.x + 10.);
                 player.kill(direction);
             }
```

src/items/galleon.rs
64kramsystem: Right :+1:

src/items/galleon.rs
Zac8668: The error happens only when this code is run, when the player collides with the flying galleon.

64kramsystem: I've hit something I believe is simlar in the past. There is a MQ undocumented feature that prevents two (scene) nodes being in scope at the same time, so that the second query would return None; in that case, using braces was used to drop the first instance before the second was used. I'm checking this - there may be something similar going on.

64kramsystem: Very interesting - it's a borrowing problem indeed.

Specifically, `Player#kill` [queries `Resources`](https://github.com/fishfight/FishFight/blob/ff14a11ecde0a255a1bac3a1f1cdfeb0b8752640/src/nodes/player.rs#L280); if when such method is invoked, a `Resources` instance is in scope, then there will be a double borrow.

The braces cause the `Resources` instance to be dropped immediately after (fxs) usage, avoiding the problem. There is at least one `Resources` design that cleanly solves this (adding helper methods like `Resources#spawn_fx()`, which encapsulates the query), but it's outside the scope of this PR.

All good with the braces :laughing:

64kramsystem (approved):
Great job :star_struck:

Issue #116, opened 2021-09-25
Generalize/Categorize items


cdsupina:
Affects #129 

While I was working on writing the section for adding an item to the game (I'm using a sniper rifle as an example), I noticed that a lot of the code I was writing for the sniper rifle was directly copying the musket code. I think that we could reduce code and lower the barrier of entry to contributing to the game by generalizing items into categories. For example, instead of having `src/items/muscet.rs` we should have `src/items/basic_gun.rs`. Then we can create a general gun struct:

```
pub struct Gun {
    pub gun_sprite: GunlikeAnimation,
    pub gun_fx_sprite: GunlikeAnimation,
    pub gun_fx: bool,

    pub bullets: i32,

    pub body: PhysicsBody,
    pub throwable: ThrowableItem,
    pub guntype: GunType,
}
```

Then we can implement contractors for all the different types of guns.

```
impl Gun {
    ...
    pub fn spawn_musket(pos: Vec2) -> HandleUntyped {
        ...
    }

    pub fn spawn_sniper(pos: Vec2) -> HandleUntyped {
        ...
    }
    
    pub fn shoot(&self, node: Handle<Muscet>, player: Handle<Player>) -> Coroutine {
        match self.gun_type {
            Sniper => {
                ...
            },
            Musket => {
                ...
            },
        }
    }

    ...
}
```

We should also generalize bullets in a similar way.

Overall this will lower the barrier of entry to contributing to the game, because new coders/contributors could start off just modifying the gun struct for their new gun with adjusted values.

erlend-sh:
Sounds good! The only consideration on my mind is the extent to which we wanna introduce this right away. Could you do this for just the gun(s) at first without touching any of the other items?

I agree this is lowering the barrier to entry once it’s the new standard everywhere. On the other hand we don’t want to be too strict about fitting everything into a neat category when a new weapon is being implemented, as I’ve pointed out on chat:

> Don’t worry too much about modularization in the first pass. Easier to sort through that in a second pass once more items are in.

I feel like there’s two essential steps to making a new item:

1. Do whatever works: duplicate code and hack it up to your heart’s content. The novelty of the item behavior is the priority.
2. Clean it up. Reduce duplication and hacky workarounds. Reducing tech debt is the priority.

This could could all happen in one PR starting in draft form, or it could be two separate ones. Of course, some items are also simple enough that they won’t need to go through a Whatever Works stage.

legendiguess:
[...]

PR #136, opened 2021-09-30
Sniper weapon tutorial

cdsupina:
I created this draft PR because I want to wait to write the section of the book until after we have settled on #116. The changes in this PR currently don't involve adding traits.

I instead generalized musket to "Gun" and added more struct values to customize how instances of a gun can work, number of bullets, shot speed, etc.

I then created a "spawn_musket" and "spawn_sniper" functions in their own item files.

Let me know what you think of this solution, I'm open to and expecting to make some changes based on your feedback.


cdsupina:
Also, the sniper is in the test level, if you want to test it out. It has 2 bullets, faster shot speed, and more recoil than the musket.

olefasting (approved):

erlend-sh (approved):

assets/levels/lev02.json
```diff
@@ -224,7 +224,7 @@
                 {
                  "height":0,
                  "id":56,
-                 "name":"muscet",
+                 "name":"musket",
```
grufkork, line 227: Best to rename it now before there's even more to do 👍

grufkork (commented):
Looks good, very little duplicate code for the gun implementations. musket.rs and sniper.rs pretty much only contains data as of now, so that's a solid foundation for moving data to JSON or RON or whatever

cdsupina:
I figured that perhaps in the future we can add traits like `Shootable` for shared behavior, but for now the sniper and musket basically have the exact same behavior, but with different values.

cdsupina:
Adding a simple gun chapter is done! You can find it in book/src/simple_gun_weapon.md.

erlend-sh (approved):

book/src/simple_gun_weapon.md
```diff
@@ -0,0 +1,212 @@
...
+    fxses: &[],
+    network_ready: true,
+},
+```
+
+If you like, you can skip to the testing section to test your new weapon, but if you haven't added a new texture or modified an existing texture, you will not be able to visually recognize your item before picking it up. For this reason, I recommend you continue to the [Texture](#texture) section next.
+
+## Texture
+
+It is important to make sure that your weapon is able to be visually distinguished between the other weapons in the game. When I added the sniper rifle to the `ITEMS` array, I copied all of the data from existing musket weapon, changing only `tiled_name` and `constructor`. To give my sniper weapon a new texture, I will also need to change the "gun" texture in the `textures` field. Currently, the "gun" texture is set to `assets/Whale/Gun(92x32).png`. I'll open this file with my pixel editor of choice, [Aseprite](https://www.aseprite.org/) ([GIMP](https://www.gimp.org/) would also work fine).
```
grufkork, line 146: Could use an explanation for what `tiled_name` does (must match name in the map)

grufkork (approved):
Looks good! I suppose it could explain things more from the ground up but on the other hand, people who venture into gamedev with Rust probably are not entry-level 👍

book/src/simple_gun_weapon.md
olefasting: 'tiled_name' should probably be changed to 'item_id' as we are moving away from Tiled. I haven't added items to the map editor yet, but when I do, I will probably have to change this, as well as create a way to define these in a data file, for reference by the editor and the scene builder

cdsupina:
> Looks good! I suppose it could explain things more from the ground up but on the other hand, people who venture into gamedev with Rust probably are not entry-level 👍

I plan on adding or linking to resources for people more entry-level at some point. For example, how to clone, make a pull request, etc.

cdsupina:
I clarified what the tiled_name field is. Merging now.

PR #177, opened 2021-10-15
Replace quad-gamepad with fishsticks

### LINKED ISSUE #122: Pressing media keys on keyboard crashes the game on Linux

### LINKED ISSUE #123: Poor responsiveness with some gamepads

potatotech:
## Why [fishsticks](https://github.com/PotatoTech/fishsticks)?
Pros: 
- Fixes #122
- Fixes #123
- Delicious

Cons:
- Not battle-tested
- Depends on SDL2, which means:
    - Not native Rust
    - Increased binary size
- Pun does not make sense to British English speakers


erlend-sh:
Could you do one of those [build time comparisons](https://github.com/fishfight/FishFight/pull/125#issuecomment-929879303) again?

Anyhow, I'm fine with merging this, as long as we are keeping in mind that this could always get reversed again as well. The long term goal should be a pure-Rust gamepad implementation.

olefasting:
> > I haven't looked deeply into how Potato solved this but if we want to minimize build times (forgive me if you have done this already, @PotatoTech) we can drop the dependency on the SDL library, in its entirety and, in stead, just pull in the controller DB, much like how glirs does it....
> 
> quad-gamepad actually already uses the SDL controller database. The reason that I wanted to replace it was because it had extremely poor responsiveness with some gamepads and was detecting non-gamepad devices as gamepads, which could cause crashes in some cases. gilrs solved these problems but introduced the wonky mappings issue. Thus, I made the switch to SDL2. If we want to reduce build times, we could fork gilrs and try to fix the mappings issue, but continuing to use quad-gamepad in its current state is not the best for compatibility.

Thanks for the clarification! I think you have done the right call here and I think we should either fork gilrs or make our own interface, whichever you judge as the best solution. I also think that we should consider separating this out to its own crate, as it matures, for the benefit of the entire rust gamedev community. If you choose this solution, I will gladly help you maintain this, as it is required, outside the bounds of fishfight....

PR #180, opened 2021-10-15
Life ring

aagonzalezm:
the life-ring inverts gravity without issues and is used like a gun despite having turtle-shell like behaviour. All the weapons were updated to flip accordingly like the fish.


erlend-sh:
> All the weapons were updated to flip accordingly like the fish.

Is this something all weapons will need to take into account going forward? Sounds like something we’d want to generalize somehow. @olefasting @grufkork thoughts?

Chosenundead15:
It might not be that necessary to worry yet as it works just like the old facing parameter, pass the parent bool on the mount function and then assign it on the function. The corresponding draw function which every gun shares from GunLikeAnimation then has the additional flip_y parameter which solves that part for every gun.

PR #225, opened 2021-11-05
Fix effect offset and refactor

legendiguess:


olefasting (commented):
Good stuff but the get_weapon_effect_position should be calculated on the weapon in stead (as an offset from the position of the weapon), as it is a property of weapon, not player. Likewise, particle effects that belong to weapon, should be spawned on a particle component on weapon, not player. Weapon should be as independent of Player as possible (currently not the case, as we pull Player into the attack coroutine, but this will change in the future) so that weapons, in theory, could be used independently of player, if we, for example, wanted to implement a turret or something like that, without having to reimplement weapons for it...

PR #231, opened 2021-11-07
Object selection

Chosenundead15:
Improvements to the `SelectObject` functionality. It can select any object in the screen plus retrieving more data from them,


olefasting:
In stead of adding `selected_id` to the context, you should in stead change the type of `selected_object_index` to `Option<String>`. This was made a `usize` because I was expecting to select from an array of items, when I added this in the very beginning of implementing the editor and I have simply forgotten to change the type.
I also disagree when it comes to the renaming of the volcano texture entries. The id's they had were much more informative, so I propose we keep them as they were.

Also, the `apply_action` method should not be made public. It is private to keep the strict flow of control, where `Editor` calls ui and tools, returning actions, that are then applied to the map or the editor node, by the editor itself.  Also, there are no reasons to do this to achieve what you want, as you call the method locally.

Otherwise, it looks good!

Chosenundead15:
the apply_action being public must be from when I had the code very unorganized and required being public to work, now is not necessary but I forgot to rollback the public. Volcano and Shark is mostly a way to make sure to retrieve the right texture without the game returning nothing as objects currenlty don't have info of the texture to use on the level editor, updating that and the json might be necessary to keep the more logical use of the naming.

Choosing a way to indentified the object was a bit confusing at the moment, as initially used the index retrieved while iterating but it didn't get anything else so it make me wonder if it was the most appropriate way to select the item for any action, and accounting for not just selected items, but if it is going to at the moment just select placed ones, the recommendation of just using the index is perfect.

olefasting:
Yeah, maybe we can create a new member of `Editor` and `EditorContext`, named `selected_map_object` (`usize`) and use that for this, and I can use `selected_object` for the toolbar I am creating that will show a list of all the available items that can be placed on the map

olefasting:
I'll close this PR as I have copied the selection code from your branch into my current working branch, as I needed object selection to proceed with my own work

PR #232, opened 2021-11-07
Add schema validation for JSON files

### LINKED ISSUE #187: Add exception for all json files in workflows and add schema validation
We should add an exception for changes to json files for all our checks on PRs and add schema validation for all files.
Currently, this only applies for maps and we will, in the case of maps, have to figure out a way to differentiate between tiled maps and non-tiled maps. This is already done in the game, with a variable in the map declarations file but if we need to add some other criteria for workflows, that is fine. We could, for example, add an inner extension to tiled maps (filename.tiled.json), if it is not possible to add parameters for deducing this from the data in the data file. If multiple schemas can be used on the same file, I can add a parameter to the internal map format to accommodate for this...

I am also aiming at finishing my work on items and all the dynamic loading of resources, either this weekend or early next week, and this will introduce a whole lot of new data files that should be validated. Once this PR is merged, I will help whoever takes this on making schema files for all of this...

orhun:
Closes #187

This PR adds the initial implementation for validating JSON files using schemas. There are a few key points:

- As mentioned in `validate_json.sh` script, schemas can be added and validated in 2 ways:
  - You can add the corresponding schema for a JSON file, such as `file.schema.json`
  - Or you add a _common_ schema file named `schema.json` in a directory and it is used for validating all of the JSON files in that directory.
- Current implementation only supports [JSON Draft-07 Schemas](https://json-schema.org/draft-07/json-schema-release-notes.html) and they can be written by hand or generated using tools like [this site](https://extendsclass.com/json-schema-validator.html).
- I only added 3 generated schema files in this PR but we should probably add more schema files for coverage. I can open an issue for tracking this when this PR is merged.
- You need [ajv-cli](https://github.com/ajv-validator/ajv-cli/) if you want to run the script on your local machine.
- `validation.yml` workflow only runs if a commit includes a change about any JSON file.

#### Workflow results

Successful run:

success

Failed run:

fail

Also, you can see the workflow run for this PR here: https://github.com/fishfight/FishFight/runs/4133154596?check_suite_focus=true

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -1,0 +1,26 @@
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "**.json"
+  pull_request:
+    paths:
+      - "**.json"
+
+name: Format validation check
+
+jobs:
+  json-validation:
+    runs-on: ubuntu-latest
+    name: Validate JSON files
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install ajv-cli
+        run: npm install -g ajv-cli
+
+      - name: Validate files
+        shell: bash
+        run: ./scripts/validate_json.sh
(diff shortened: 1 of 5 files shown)


olefasting:
Awesome!

Issue #224, opened 2021-11-04
Animation implementation


olefasting:
We should make our own implementation of animations that support non-looping animations, so that we won't have to hand write what is, in essence, an animation player, in a coroutine, every time we want to play a non-looping animation.
This is very easy to implement and we can just copy-paste macroquad's implementation and add an `is_looping` member to the `Animation` struct and stop the animation on the last frame, if this is true. I don't think this will be added to macroquad in the near future (I have made an issue for it but have seen no activity on it). I reckon this is because it is a breaking change, as the `Animation` struct is constructed without a constructor, so such a change would break everything that uses Macroquad's animation system, as they would have to add `is_looping` to all `Animation` declarations, after such a change.

Anyone taking this on, should just implement this in the `AnimationPlayer` component, replacing the current macroquad animated sprite member already there. We already have a type that replaces macroquad's `Animation`, so as mentioned above, it is just a matter of copy-pasting the update and draw methods from macroquad over to the `AnimationPlayer` component and add a check for whether or not the current frame is the last frame, if `is_looping` is `true`...

tigleym:
Looking for a good first issue! Mind if I give this one a shot?

olefasting:
> Looking for a good first issue! Mind if I give this one a shot?

That would be very helpful, so go ahead!

PR #235, opened 2021-11-09
Add support for non-looping animations

tigleym:
For #224. This largely copies over macroquad's `update` implementation for animations, but instead the animation can be halted when an animation is configured to not loop. We still rely on macroquad's `AnimatedSprite` implementation to contain the current animation state, but it seemed necessary to track `time` and the `current_frame` on `AnimationPlayer` itself since those fields aren't exposed.


olefasting:
Awesome!

PR #236, opened 2021-11-09
Reorder book summary

erlend-sh:


tigleym (approved):
LGTM! Thanks!

erlend-sh:
@orhun could this summary file be excerpted from future build checks? It's not like like the other .md files.

tigleym:
I suspect this action is causing the checks to not go through: ["Create and Publish Github Release"](https://github.com/fishfight/FishFight/runs/4152602029?check_suite_focus=true). Interestingly, it also happens on my [own fork of the repo](https://github.com/tigleym/FishFight/actions). 

In particular: this error `Error: ⚠️ GitHub Releases requires a tag`. I don't totally understand why a local PR would trigger this action vs. one from a forked repo (at least this is what I've observed from the PRs with successful checks here).

orhun:
> I suspect this action is causing the checks to not go through: ["Create and Publish Github Release"](https://github.com/fishfight/FishFight/runs/4152602029?check_suite_focus=true). Interestingly, it also happens on my [own fork of the repo](https://github.com/tigleym/FishFight/actions) (and orhun's it seems).
> 
> In particular: this error `Error: ⚠️ GitHub Releases requires a tag`. I don't totally understand why a local PR would trigger this action vs. one from a forked repo (at least this is what I've observed from the PRs with successful checks here).
> 
> Some digging around I found this: [softprops/action-gh-release#20](https://github.com/softprops/action-gh-release/issues/20)

Yes, I realized this as well. Pretty weird behavior. I don't know why it runs "on its own" when there is not tag created either.

Maybe we should consider switching to a more stable action for uploading release assets, such as [upload-release-action](https://github.com/svenstaro/upload-release-action).

tigleym:
> Yes, I noticed this as well. Pretty weird behavior. I don't know why it runs "on its own" when there is not tag created either.

I read that `create` will also trigger workflows for newly created branches: https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#create. I'm not sure if not specifying a branch here is causing the workflow to run when any new branch is created. 

Maybe we could do something like:

```yaml
on:
  create:
    branches:
      - 'release*'
    tags:
      - 'v*'
```

Another thing we could try is instead of running on `create` we could do it on `push`?

> Maybe we should consider switching to a more stable action for uploading release assets, such as [upload-release-action](https://github.com/svenstaro/upload-release-action).
> 
 Definitely worth looking into!

orhun:
> I read that `create` will also trigger workflows for newly created branches: [docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#create](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#create). I'm not sure if not specifying a branch here is causing the workflow to run when any new branch is created.

Exactly! Good catch 🐟 
 
> Another thing we could try is instead of running on `create` we could do it on `push`?

That's definitely going to solve the "running on its own" issue I think. I always used `push` for CD workflows in my projects and every time it worked fine. I totally overlooked that `release` workflow was using `create` in this case 😦 

> > Maybe we should consider switching to a more stable action for uploading release assets, such as [upload-release-action](https://github.com/svenstaro/upload-release-action).
> 
> Definitely worth looking into!

Can I make this changes and submit a PR if you're not already working on this? :bear:

tigleym:
> Can I make this changes and submit a PR if you're not already working on this? :bear:

I haven't worked on it. So it would be great if you could do that! 🙏

erlend-sh:
@orhun could we temporarily disable whatever check is blocking this just to get it merged?

orhun:
> @orhun could we temporarily disable whatever check is blocking this just to get it merged?

Checks in the actual repository are not being run for pull requests. Only the workflows in your forked repository (at the time of PR commit - 0d4a515) are checked. So you can disable the workflows by updating your `erlend-sh-patch-1` branch. But, I don't think that's the reason why we can't merge because there seems to be a merge conflict.

I can solve the merge conflict and try to merge the PR for you, if you want 🐻

erlend-sh:
> I can solve the merge conflict and try to merge the PR for you, if you want 🐻

Please do 🙏

orhun:
I resolved the conflict, but still cannot merge because of this message:

Screenshot 2021-11-15 at 22-24-05

Then I decided to go into repository settings and saw that we have the `main` branch protected and it was requiring some status checks to be passed before merge:

Screenshot 2021-11-15 at 22-23-24 Branch protection rule

But the problem is, we don't run these checks for `*.md` files. So we're stuck in a dilemma 🙂 

Should I remove those settings or temporarily disable the exclusion of markdown files? (I'd go with 1️⃣ since this might cause other problems in the future) What do you think? @erlend-sh

erlend-sh:
1️⃣ 👍

PR #254, opened 2021-11-16
Juice book

grufkork:
First version of the juicing page. Couldn't figure out much more to write regarding the particles without basically writing docs for the particle system (though it could use some). Are there some more sections that should be covered? Should I document things such as the screen shake system, like with actual code?


erlend-sh:
> Should I document things such as the screen shake system, like with actual code?

That’d be good, yes 👍

olefasting:
I have made some changes to the particle system in #262 that should simplify it a bit. Basically, you will just add any number of `ParticleController` component to types that should be emitting particles, or a set of `ParticleControllerParams` when defining things like weapons and passive/active effects in JSON.

grufkork:
Alright, I think I've figured out how the particles work? 

Also I sense issues keeping this in sync as the code updates. How's the prospects regarding generating docs, like many crates has? If it can be done only for the backend/API parts so it doesn't get cluttered.

olefasting:
Yeah, the API isn't super stable at this point, especially not when it comes to items. Although it is at a state where it can be used to implement most of the things we want and without any (known) bugs, I am constantly tweaking it as I go; especially the parts that are most relevant for docs, like the JSON data models. 

When this is done, it will probably just be a matter of documenting a few interfaces to cover most things, much like how it has been done in the editor.

grufkork:
All right then. Is there anything more you all would like me to cover, or further elaborate?

erlend-sh:
Looks great! 👍

Further iteration should be based on questions that come up from learners. LGTM.

book/src/SUMMARY.md
```diff
@@ -10,7 +10,11 @@
...
   - [Environmental Weapons](./environmental_weapons.md)
 - [Netcode](./netcode.md)
+- [Juice](./juice.md)
 - [Editor](./editor.md)
 - [Contributing](./contributing.md)
   - [Contributing Code](./contributing_code.md)
   - [Simple Gun Weapon](./simple_gun_weapon.md)
+- Documentation
+  - [Screen Shake](./docs/screenshake.md)
+  - [Particles](./docs/particles.md)
```
tigleym, line 20: I feel like these should be sub-sections of the "Juice" chapter? Although I can see the need to have a separate chapter focusing on implementation details, which we may want to revisit at another point.

book/src/juice.md
```diff
@@ -0,0 +1,43 @@
+# Juice and Feel
+**Juice**, or **game feel**, is the science and art of making pushing a button viscerally satisfying. Well, more precisely it's about giving feedback to confirm the player's actions. For games where flow and feeling matter, juice is the thing that makes the player understand your game without reading dials and processing numbers. Juice gives a game life and soul that makes the player feel like a part of the world. Juice is what makes a game feel fun without actually making any mechanical difference. And juice might just be what makes or breaks your game.
```
tigleym, line 2: Neat! I didn't know this.

book/src/juice.md
```diff
@@ -0,0 +1,43 @@
...
+
+In Fish Fight, [macroquad-particles](https://crates.io/crates/macroquad-particles) submodule is used to handle particles. While it might look rudimentary, it is deceptively so. With several particles, blending modes and curse pretty impressive effects can be made. The explosions in FishFight used to be entirely procedural! Now it is used for muzzle smoke, trails, bubbles and whatever other details will be added in the future. 
+
+Particle presets are stored in .json files, which can be either generated manually or using [this excellent editor](https://fedorgames.itch.io/macroquad-particles) which provides previews and live tweaking. It might not be entirely up to date with the latest particle system format, but should get you up and running quickly.
```
tigleym, line 43: I really like this section. I wasn't sure how we were finding those values in the particle json files, so this helps me understand this part of the codebase a bit more.

tigleym (approved):
This looks good to me! Now I know what "juice" is in this context. I had one comment about the organization in SUMMARY.md but I don't think it should block landing this.

book/src/SUMMARY.md
grufkork: Oh yeah, I was going to ask about that... The "docs" are like in between code documentation and a guide, but aren't strictly related to juicing? I'm thinking we can leave it like that at the moment and find some better structure once more text drops in, it might be more apparent then.

book/src/juice.md
grufkork: It only took a bunch of "find all references" clicks and a while... I can't guarantee it's right but the others are saying lgtm 👍

Issue #152, opened 2021-10-03
[SPEC] Blunderbuss & BlunderBass


SirGiraffey:
![blunderbass and blunderbass](https://user-images.githubusercontent.com/57463655/135750130-6c3d0bab-7093-45b8-b842-fe6b8abcae07.png)

erlend-sh:
We've got the art completed for this by @BumbertFiddlesticks:

#### Blunderbuss (remix of gun)
![image](https://user-images.githubusercontent.com/583842/140936896-c2a66c47-9ca4-4eef-b41d-468941938dd4.png)

#### Blunderbass

![image](https://user-images.githubusercontent.com/583842/140936946-105e958e-d4a9-4b24-932a-6672a4af5ad0.png)

Only the coding part of the task remains.

SirGiraffey:
Love the design, although the stock and the fish tail don't look right. Maybe, as you said, remove the tail altogether or turn the tail into the stock of the gun

billyb2:
If it's alright, I'd like to try to work on this issue

BumbertFiddlesticks:
I'm not working on this anymore, but the blunderbass above isn't the final version! Not sure how .zip friendly github is but you can see final here: [Blunderbuss +BlunderBass.zip](https://github.com/fishfight/FishFight/files/7571727/Blunderbuss.%2BBlunderBass.zip)

If you still want to improve it, that's fine too of course, and best of luck with the project :)

billyb2:
@BumbertFiddlesticks the art looks great! Thank you very much!!!

PR #270, opened 2021-11-22
Added the BlunderBuss

billyb2:
As defined in #152. Shoots 5 bullets with moderate recoil, and 2 seconds between 2 shots. I do recognize that I need to fix the offset and the actual BlunderBuss projectiles, but overall I think it's a good start. Still a WIP.

https://user-images.githubusercontent.com/45850508/142883242-775b4343-0284-4a2f-93f7-b27282699e86.mp4

https://user-images.githubusercontent.com/45850508/142893624-0ae07e22-53c9-40fc-911b-be4775cbb15d.mp4

- [ ] Blunderbuss projectile sprite
- [x] Blunderbuss offset
- [x] Blunderbass
- [ ] Blunderbass projectile sprite
- [x] Blunderbass offset


billyb2:
I think that the PR is ready for review, since the projectile sprite stuff needs to be discussed first. Personally, I think that the bullets it uses now look good.

PR #316, opened 2022-01-04
Have in-game credits AnonymousPro-Regular font instead

tigleym:
The internal font lacked support for the "dotless i" character, which doesn't look too great. This PR instead uses another font: AnonymousPro-Regular.


olefasting (approved):

olefasting:
Unsure why this gives compilation error now (didn't do so before merge). Could you please check it out and submit a fix, if possible?

PR #318, opened 2022-01-04
Update CREDITS

erlend-sh:


erlend-sh:
Jacob wanted to include his artist title, Emerald Jak.

@SirGiraffey wanted to include his full name: Michał Grzegórzek - Kiaszewicz

Not entirely sure how to deal with the dash, since we’re already using that for separation.

SirGiraffey:
If the dash is a problem you can remove it, no problem

CREDITS.md
```diff
@@ -18,7 +18,7 @@
 - Carlo Supina, [@cdsupina](https://github.com/cdsupina) – Dev Advocacy
 - Micah, [@tigleym](https://github.com/tigleym) – Dev Advocacy
 - Drake, [@Drakeyn](https://github.com/Drakeyn)– Media Content
-- Zyrafa, [@SirGiraffey](https://github.com/SirGiraffey) – Gameplay Design
+- Michał Grzegórzek - Kiaszewicz, [@SirGiraffey](https://github.com/SirGiraffey) – Gameplay Design
```
tigleym, line 21: It looks like the 'ł' character is also missing from the current font set we're using. #316 introduces a new font for the credits this should be fine once it's merged.

tigleym (approved):
LGTM! Quickly testing this and it looks like we'll have to adjust the x-offset a bit to accomodate smaller window sizes (the text end up getting clipped off screen for longer labels). But I think that can be addressed in another PR.

PR #326, opened 2022-01-11
ECS refactor

olefasting:
This is a draft PR of a refactor, using hecs in stead of Macroquad scene.

Firstly, sorry for the delay, guys.

The merging process shed light on a few problems that I needed to resolve. I had also forgot to implement the sproinger...

I have integrated all the juicing changes from @legendiguess, so it should be up to date with the state of the master branch. If I missed something or if something is in other ways not as it should, let me know and I will correct it :)

Please note that the current systems, especially those related to Player, were created with a focus on code reuse. This means that there are a few giant systems that will be cleaned up at a later stage. I am primarily thinking of `update_player_state` and `update_player_inventory`, which should both be split up into more appropriate, smaller systems. Using hecs is way more flexible than scenes, though, so it won't be that problematic to clean this up as we go...

We no longer need to use coroutines now and I have therefore removed the usage of state machines. For now, player state is just a struct of `bool` flags, for example (this made more sense than using an enum as not all states are mutually exclusive).

There has also been some minor changes to the JSON API, primarily around items. This is mostly renaming to make the API more intuitive, like renaming `animation` to `sprite`, to avoid confusion with its member array `animations`. I have also added drop and depletion behavior for items and weapons.

The base `Item` struct has also been removed and this name is now used for what was previously called `EquippedItem`. This means that a `MapObject` can be a piece of `Decoration`, an `EnvironmentItem` (sproinger), an `Item` or a `Weapon`.

I have also removed the sprite that we previously used for items that were on the ground as we now just use an animation with the id `ground`, or `idle` if this has not been defined, in stead. I have also removed all use of delta time from all the physics calculations as this is not needed when we use fixed updates. This means that when you define force, speed and similar values, you will have to divide it by 120 (I believe this is the number of fixed updates per second), compared to how it was before.

Decoration has also been implemented properly now and they can be defined in `decoration.json`.

In the future we will change the API further, in order to better reflect the internal API. When we implement modding, we will be defining components (pure data types) internally or in JSON and create systems (logic acting on components) either in rust or by using whichever solution we end up integrating for script support. I therefore think we should design the JSON API to reflect this, so an item, for example, would just be a JSON array of component definitions, used as a blueprint when spawning the item.

EDIT: I forgot to add that I also need to resolve an issue with draw order. `Sprite` and `AnimatedSprite` are components in their own right, so weapons are their own entities now (`PhysicsBody` is just disabled when picked up and position set in the `update_player_inventory` system). This means that weapons, being spawned after players, will be drawn over all players all the time, in stead of being drawn at a z-depth relative to the player that is carrying it. In other words, if two players pass each other, both carrying weapons, the one carried by the player that is drawn behind the other, will still be drawn on top of the player that is drawn on top of him.
I will fix this by adding a z-depth variable to `AnimatedSprite` which is set in the `update_player_inventory` system, and just sort the query used by the `draw_animated_sprites` system by this value. I'll probably add that before this is getting merged, so no need to make an issue about this if you decide to try this version out :)


grufkork:
I'm getting an error:
```
error: linking with `link.exe` failed: exit code: 1181
[...entire compilation command...]
 = note: LINK : fatal error LNK1181: cannot open input file 'SDL2.lib'
```
I see I need to get SDL dev stuff or something, but why has this not been needed before?

src/player/inventory.rs
```diff
@@ -0,0 +1,408 @@
+use std::os::linux::raw::stat;
```
grufkork, line 1: Is this used? It fails on mac and win, and I can't find any references to it

legendiguess:
> I have integrated all the juicing changes from @legendiguess, so it should be up to date with the state of the master branch. If I missed something or if something is in other ways not as it should, let me know and I will correct it :)

Thanks 😄

src/player/inventory.rs
olefasting: No, that is probably an error due to autoimport. I have probably pressed enter too early when writing `state`...
I am doing a cleanup with clippy right now and will be removing all the unused imports, like this one

olefasting:
> I'm getting an error:
> 
> ```
> error: linking with `link.exe` failed: exit code: 1181
> [...entire link command...]
>  = note: LINK : fatal error LNK1181: cannot open input file 'SDL2.lib'
> ```
> 
> I see I need to get SDL dev stuff or something, but why has this not been needed before?

We have disabled embedded SDL to resolve the build issues we had, so you now need the library on your machine in order to build

grufkork:
I see, found #324. I'll try adding the feature conditionally

src/error.rs
```diff
@@ -26,18 +26,26 @@ struct Custom {
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum ErrorKind {
     General,
+    EcsError,
```
PotatoTech, line 29: This variant is the only one with an "Error" suffix. Since the enum's name is already `ErrorKind`, this can be left out.

olefasting: Yeah, they all had suffixes. so I must have forgotten to remove it

orhun:
I haven't looked at the code in depth (oh boy, a lot of changes 😄 ) but there are a few things that I realized at first glance about the gameplay. (I'm not sure they are trivial issues or not.)

- floating property seems to be gone. I can't see my fish float on the air when I press jump mid-air:

<details><summary>GIF</summary>

rec_20220111T225621

</details>

- jump points feel so OP. (they make you jump so high):


<details><summary>GIF</summary>

rec_20220111T225621

</details>
rec_20220111T225749

- weapon throw should be improved to throw the weapon further than the player. I can press throw repeatedly and it feels like I've never thrown the weapon:

rec_20220111T225853

grufkork:
I am going to do some work on weapon throwing, making it a more core mechanic, so you can probably hold off on that for now 👍

grufkork:
> I see, found #324. I'll try adding the feature conditionally

...turns out cfg()-stuff dos not work for enabling features? :/ 
Found [this](https://github.com/rust-lang/cargo/issues/1197) but it's unresolved. 

Testing locally:
```toml
[target.'cfg(windows)'.dependencies]
fishsticks = {version = "0.2.0", default-features = false} #bundled-sdl2 is not a default feature, just being sure

[target.'cfg(unix)'.dependencies]
fishsticks = {version = "0.2.0", default-features = false, features = ["bundled-sdl2"]}
```
It still compiles on Windows, fails only if the features statement is completely removed. So that's that :P

olefasting:
>     * floating property seems to be gone. I can't see my fish float on the air when I press jump mid-air.

I had just forgotten to apply the floating factor to the gravity. Fixed it now, in my last commit

>     * jump points feel so OP. (they make you jump so high)

Yeah, jump needs a little work. I removed the old jump, which was a bit better than this, as I don't think it would work well for network play. Right now, it just applies a very high amount of force when you press jump, which is immediately counteracted by gravity. This creates very jerky jump and it is very hard to find the right combination of jump force and gravity, without it either being like this or gravity being way too low.
The plan is to create a system for fading gravity in after a jump is started, making the jump curve more smooth.
Things like move speed also needs to be tweaked a bit. I just divided all forces by 100 after removing multiplication by delta time but I think it should in stead be 120, as that is most likely the actual rate (delta time being 1.0 / 120.0)

grufkork:
I'm having a strange issue where using items freezes the game for a split second... The sniper, both kinds of bombs and musket cause this. Sword, MG, mines and cannon does not. I'm on a low-end laptop right now but everything else runs smoothly, including explosions and other FX. 

Looking at the JSON files, it might have to do with particles. Those that cause it have a "particles":[...] statement while those that don't freeze have a "particle_effect" under the effects list.

olefasting:
> I'm having a strange issue where using items freezes the game for a split second... The sniper, both kinds of bombs and musket cause this. Sword, MG, mines and cannon does not. I'm on a low-end laptop right now but everything else runs smoothly, including explosions and other FX.
> 
> Looking at the JSON files, it might have to do with particles. Those that cause it have a "particles":[...] statement while those that don't freeze have a "particle_effect" under the effects list.

Are you sure it is not just the attack duration taking effect? The player is "stun locked" for `attack_duration` (`state.is_attacking` is `true`), which is also when recoil is applied. An undefined or zero `attack_duration`, as is the case for sword, will only restrict attack interval by `cooldown`, not locking the player and apply any recoil...

grufkork:
I'm pretty sure it isn't the stun period, nothing seems to move. I tried the current main branch, it's a bit different... Cannon causes some lag (note that the cannon has some particle FX in pre-ECS version).  
Bombs in ECS freeze big time when throwing, none when exploding. Bombs in pre-ECS freeze a short while when throwing and even longer when exploding.
Sniper freezes for a moment, but after a couple of shots it runs smoothly. Almost sounds like some kind of caching issue? I believe the particle systems are created with the item, not when activated, so it should not be that either.

olefasting:
> I'm pretty sure it isn't the stun period, nothing seems to move. I tried the current main branch, it's a bit different... Cannon causes some lag (note that the cannon has some particle FX in pre-ECS version). Bombs in ECS freeze big time when throwing, none when exploding. Bombs in pre-ECS freeze a short while when throwing and even longer when exploding. Sniper freezes for a moment, but after a couple of shots it runs smoothly. Almost sounds like some kind of caching issue? I believe the particle systems are created with the item, not when activated, so it should not be that either.

I don't think the cannon had any particles; the effect is just a secondary animation player with a smoke ring sprite. Might it be a draw issue..?
Any ways, we will have to keep a lookout for similar issues or a way to reproduce it...

grufkork:
> The plan is to create a system for fading gravity in after a jump is started, making the jump curve more smooth.

A nice thing to do is to lower/disable gravity while up is held, allowing for finer control over jump height. I think gravity is a tad high too, with lower vertical movement speed it's easier to predict and follow your and other player's trajectories.

olefasting:
> > The plan is to create a system for fading gravity in after a jump is started, making the jump curve more smooth.
> 
> A nice thing to do is to lower/disable gravity while up is held, allowing for finer control over jump height. I think gravity is a tad high too, with lower vertical movement speed it's easier to predict and follow your and other player's trajectories.

That is what we used to have. As we are implementing networking I would prefer a solution that could be implemented as a one off action, server side, without requiring continuous input. Mostly, because we probably want to tweak jump, so that we have it exactly high enough to reach a height of _n_ tiles and a distance of _n_ tiles. This makes it possible to make challenges to reach certain areas in maps, etc, requiring timed jumps. If you have a few percent packet loss maybe one in between twenty and fifty such jumps will fail, due to missing jump input. That is maybe not a huge problem, though, but it should also be possible to emulate the same behavior by just LERPing gravity in and out, deterministically...

erlend-sh:
Our base jump mechanics should follow Duck Game very closely. Networked play needs to adapt to that, not the other way around.

> As we are implementing networking I would prefer a solution that could be implemented as a one off action, server side, without requiring continuous input.

We can do this, but it should be layered on top rather than being the default core mechanic.

olefasting:
Not sure how it works in duck game but I can have a look...

PR #331, opened 2022-01-21
Update Cargo.toml & remove unused dependencies

orhun:
This PR makes the following changes in `Cargo.toml`:

- `license` and `description` sections are added.
- `authors` section is updated to something more general since we have `CREDITS.md` now.
- Unused dependencies are removed.
- Formatting: reordered the dependencies/sections & removed newlines.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,12 +26,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anyhow"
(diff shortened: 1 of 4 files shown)


olefasting:
bincode will be used for binary encoding in the networking pr that is coming, that's why its there, but I'll just re-add it when it's needed :)

PR #364, opened 2022-02-04
Cannon juicing

legendiguess:
This PR is part of https://github.com/fishfight/FishFight/issues/120


olefasting:
What's the deal with the parameterization of gravity? This seems like something that should be solved by turning graity off on the body and adding force, in stead

PR #381, opened 2022-02-11
Hats!

olefasting:
This adds hats to the game. Lots of hat graphics has been added to `assets/textures/items/hats/` (courtesy of **zTecna**) and one example implementation has been added at `assets/items/pirate_hat.json`.

Hats are defined exactly like a normal item, only you also set `hat` to `true` in the item definition json file when defining a hat


src/player/controller.rs
```diff
@@ -88,8 +84,8 @@ pub fn update_player_controllers(world: &mut World) {
                 let input = collect_local_input(*input_scheme);
                 controller.apply_input(input);
             }
-            PlayerControllerKind::Network(_account_id) => {
-                // TODO: Network input
+            PlayerControllerKind::Network(_player_id) => {
+                // not implemented
```
orhun, line 88: How about using [` unimplemented!();`](https://doc.rust-lang.org/std/macro.unimplemented.html)? 🐻

olefasting: nah, I am currently working on that part, I just had to remove some code so I wouldn't have to branch to add the has...
It won't be triggered any ways, so no need for a panic there

PR #403, opened 2022-03-08
Engine backend API

olefasting:
This is a preview of the upcoming changes, as we move Macroquad behind a backend abstraction layer in core, so that we can use an in-house renderer with the ultimate release.

The renderer is not yet implemented (it is the next thing on the program, so expect that some time tomorrow or on Thursday). There is also a lot of stuff that isn't wired in yet for the internal backend, but using the Macroquad backend now works as it should. Some of the refactoring has been done a bit hastily, as we expect to transition our entire code base to our internal backend, as soon as possible (some messy conversions between Sizes and Vec2s, here and there, for example).
As mentioned, a lot is still missing from the internal backend, but aside from the renderer, which I will implement over the coming days, using OpenGL, most of the things can be copied from Macroquads code base (file handling, for example)


orhun:
Woo, `macros`! Fancy! 🍫

olefasting:
> Woo, `macro`! Fancy! chocolate_bar

Yeah, I added it so that it is possible to also apply mods to resources implemenented outside the engine :)

olefasting:
> 

We might consider this in the future but for now we are just going with a standard OpenGL renderer

erlend-sh:
On a different note, [ARCHITECTURE.md](https://github.com/fishfight/FishFight/blob/main/ARCHITECTURE.md) will surely need an update to go along with this PR.

olefasting:
closing this to reopen a new one

Issue #149, opened 2021-10-02
Poseidon's Trident


SirGiraffey:
![Poseidons Trident](https://user-images.githubusercontent.com/57463655/135719529-7f856d4a-57f8-4fee-afed-0045594b9df7.png)

erlend-sh:
Let’s have all kinds of Poseidon’s tridents. Maybe this first one is the Blue Trident.

Then we can have the Red Trident which is a death ray, and a Fake Red Trident (with subtle white stripes and faintly different charge sound) which self-destructs and explodes.

etc.

Anyone who wants to spec out another trident should make a dedicated issue for whichever color they’ve picked.

erlend-sh:
Have at it! We’ve got an initial iteration of the trident art asset completed as well, I’ll send you the zip on Discord.

![DF300F0F-0A07-4BD3-9808-428E23C4CBF7](https://user-images.githubusercontent.com/583842/138313067-759e56d1-f201-46d0-b9f0-9f2ecc524aff.png)

PR #409, opened 2022-04-05
Add Item: Trident

zicklag:
Hey folks! I got the assignment from @erlend-sh to make at least 2 weapons, so here's the first one. ( Thanks to @tekhnaeraav for the idea to make a trident ).

The sprite is just a placeholder, of course.

https://user-images.githubusercontent.com/25393315/161655054-e7901ee5-69c7-474e-8301-06f46443473a.mp4

I'm not 100% sure how you guys want to balance the weapons, but the trident currently swings faster than a sword, and the damage zone doesn't cover your own body, just the area in front, so it's got more of a spear behavior than the sword.

I think it measures more powerful than the sword in general, but still probably worse than most guns. 

Also, I noticed that when laying on the ground and facing to the left, the sprites bounding box isn't quite right:

image

image

It turns out that the other weapons for the most part don't have a lot of empty space in the sprites on the x axis, so it's almost un-noticeable, but the trident has a lot of extra horizontal room for the stab animation so it get's harder to pick up the trident when it's facing left.

I'm assuming that should be tackled as a separate PR/issue, but if you want I can look into fixing it.


zicklag:
I just realized that you guys already had a set of item specifications to implement. Whoops. :upside_down_face: 

Well there's this one if you want it, but I'll work on a couple of the ones from your list, too. :slightly_smiling_face:

erlend-sh:
Whoops, I forgot to link to the standard list of [Items](https://github.com/fishfight/FishFight/labels/item).

However, I don’t think we’ve implemented #149 so you can go ahead with that as well.

PR #411, opened 2022-04-06
Refactor Passive Effects

zicklag:
Hi @olefasting, I sent you an Email asking about how I should go about adding new passive effects such as a "controls reversed" effect, and I started experimenting with a refactor of the passive effect system along the lines of what I was thinking.

I moved passive effects to an enum so that each one can have it's own config options. So shield has back and front options, but a `speed_multiplier` effect could have `x` and `y` options.

Depending on how soon you get to reviewing this, I'm just going to be experimenting with the design, adding more effects and testing it with weapons to see how the API works out.

Also, I know I removed a lot of code, which was kind of presumptuous, I just don't know exactly what the full goal is, so I'm experimenting to get a feel for things until you let me know what direction we want to go in.


zicklag:
Changing strategy in response to Email.

PR #414, opened 2022-04-08
Passive Effect Sprites

zicklag:
Depends on #413.

Makes an easy way to add sprites to passive effects.


zicklag:
Clippy is complaining because adding a `sprite` field to the passive effect makes it's variant much larger than the other in `AnyEffectParams`. It suggested boxing `AnimatedSpriteMetadata`, is that something we want to do?

orhun:
> Clippy is complaining because adding a `sprite` field to the passive effect makes it's variant much larger than the other in `AnyEffectParams`. It suggested boxing `PassiveEffectMetadata`, is that something we want to do?

Sure, I think it is fine to Box it.

Issue #337, opened 2022-01-24
Critters


SirGiraffey:
Critters would be small entities with relatively simple AI and animations. They wouldn't affect the gameplay in any meaningful way, but  they would make the game feel much more organic.

Small Fish: They would float around the level randomly, maybe form clumps and wouldn't drift too far from each other. When a player gets too close they get scared and swim away. When a bullet passes closely, they could panic and swim away quickly, completely disorganizing the "group". If hit by a laser or a bullet, they could turn into a tiny skeleton (or sushi for family-friendliness) that drops to the ground.

Crabs: They would walk left-right randomly and stop for a few moments, but they wouldn't walk off ledges on their own. If a player gets too close they would walk away, and drop from platforms if there is no wall to stop them from doing so.

![obraz](https://user-images.githubusercontent.com/57463655/150713079-dda5dfdb-d1bc-4094-84d8-744cd537859c.png)
![obraz](https://user-images.githubusercontent.com/57463655/150713084-1f2446cc-1037-4586-a290-7b4b86434987.png)

erlend-sh:
This is for sure happening 👍 Might also be related to #43 

And I would welcome more impactful gameplay affectations as well. These sorts of incremental embellishments are how we add unique character to our game and ensure that although it’s a spiritual successor, it has its own soul 💫

SirGiraffey:
Just made a mock-up of another critter which could serve some gameplay purpose, but mainly be a great ambient feature, the angler fish. It glows in the dark!
![Critters 3](https://user-images.githubusercontent.com/57463655/156923979-c915e4ca-bcb9-4576-8b88-cef242ae6a1d.png)

erlend-sh:
Update from @zTecna:

![819CFD6A-249A-4FB7-8CE0-0F07A7466A61](https://user-images.githubusercontent.com/583842/161223438-01b7d3c2-85e7-496a-a500-fac0e26611de.jpeg)

zicklag:
I'm going to be working on this. If anybody could upload some of the raw sprites for the crabs and little fish that would help, but nothing's stopping me from just using some placeholder sprites, so it's not a big deal.

Also @erlend-sh I wanted double-check on the scope of the MVP. I'm thinking:

- We have crabs:
  - that can be placed on the map to define their spawn point
  - that can walk left and right randomly within a small distance
  - that will try to stay a minimum distance away from players by walking away from them
  - that, for the sake of scope for now, won't have any logic to avoid ledges
[...]

PR #417, opened 2022-04-10
Crabs! 🦀

zicklag:
I made a rough copy of the crab sprite from #337 to use for them, but the shading isn't as nice so we can update the sprite later.

https://user-images.githubusercontent.com/25393315/162595716-d0645cb5-d411-4b70-9584-99c57f29af31.mp4


src/utils.rs
```diff
@@ -0,0 +1,34 @@
+#[derive(Debug)]
```
orhun, line 1: I think it would be a good idea to rename this file to `timer.rs` or move it to e.g. `utils/timer.rs`. "utils" is a very general name and we might want to add more "helper types" like this in the future so it is better to draw the lines now I think 🙂

Also, just a hint, this module looks so unit-testable 🐻

zicklag: Sounds good, I pushed an update.

orhun (approved):

PR #418, opened 2022-04-16
Fish School Critters and Improved Crab Sprites.

### LINKED ISSUE #337: Critters

zicklag:
I believe this closes #337 .

I don't think we want the small fish included in the schools yet, so I commented out the code that adds them and I left their sprites.

I can strip those out of the PR if you want.

---

@zTecna I set the creator of all the sprites with your artwork in it to `https://github.com/zTecna` so that you're credited. We'll probably want an `artwork.md` to track art contributions or something at some point, but for now you're in the metadata at least.


zTecna:
I just saw the fish fight twitter post! I loved how well they were applied!!

PR #422, opened 2022-05-10
Item starfish

### LINKED ISSUE #153: [SPEC] Throwable Starfish
Throwable starfish

zicklag:
Resolves #153.


zicklag:
OK, just finished both starfish and starfish launchers!

Along the way it added some useful new options to the item JSON options:

- Effects can spawn items onto the course
  - This could be useful in the future if we wanted like a banana gun to leave banana peels behind
- Effects can have a timed expiry at which they may trigger different effects than if they were triggered
- Effects can be spawned with a spread specified for their velocity, similar to projectitles
- Effects can spin in mid-air
- Effect can  be grabbed, with a customizable grab region

orhun (approved):

PR #462, opened 2022-08-02
Rename fight to folk

erlend-sh:


erlend-sh:
This is not complete, and there are typos 🤣 that’s as far as I’ll get on a tablet though.

orhun:
done in #463

PR #482, opened 2022-11-23
Attempt to Fix Multiplayer Desync

zicklag:
This is based on the guess that the desync came from the lack of determinism in the scenario when you grab an item while touching multiple items, and it isn't necessarily specified which one you pick up.


zicklag:
bors merge

PR #490, opened 2022-11-29
Implement Grenades ( + Other Small Tweaks )

zicklag:
Implements grenades and some other small tweaks.

Almost ready, just need to fix using grenades when in rollback mode.


zicklag:
bors merge

PR #520, opened 2022-12-13
[Merged by Bors] - feat: add micro-ECS crate for future use in core game loop

zicklag:
This is just my branch where I'm messing around with an implementation of https://github.com/fishfolk/jumpy/discussions/510.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       fail-fast: false
       matrix:
         config:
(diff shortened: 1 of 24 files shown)


zicklag:
I managed to get a first draft of the ECS storage rewrite.

I implements the same API as Planck, but builds it on top of an un-typed storage layer of pure bytes. This allows us to be more flexible for things like mods.

zicklag:
There's still work to do, but I'm going to call this ready to merge. We've got miri tests and 76% code coverage, which is pretty good! I've commented almost all the unsafe code.

I think it's a good start.

bors merge

PR #527, opened 2022-12-17
Change Grenade Mechanics

zicklag:
- fire-action should light it, not throw it.
- throw with drop-button. Gets thrown further based on character momentum.
- possible to pick up grenade again after it’s been thrown (by whoever).


zicklag:
bors merge

PR #528, opened 2022-12-17
Update Links In Readme From fishfolks to fishfolk

zicklag:


zicklag:
bors merge

orhun:
Hey @zicklag, just wanted to mention that it would be nice if you use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), e.g.

Instead of: 

```
Update Links In Readme From fishfolks to fishfolk
```

You can use:

```
docs(readme): update links from fishfolks to fishfolk
```

It's just better for everyone 🙂

zicklag:
Yeah, I'll look into this. 👍

My only hesitation is that I don't want to force contributors to think about it, and because we use Bors for merging, I need a way to make sure that I can set the commit messages before merging, but I can't use the GitHub squash merge feature.

I think there's a way to make this work, though, because I can have bors squash commits before merging, and I think there's a bors feature for setting the commit title from a template.

zicklag:
Ah, it looks like bors will use the PR title if you use squash merging. The only caveat is that the PR number ends up in parenthesis at the end of the PR title.

Is having the PR number in the short description a concern?

orhun:
Don't think so.

zicklag:
Thanks for pushing me to do this, now we've got a very nice setup!

I've got CI making sure that the PR title matches the conventional commit specification, and I've also configured bors to make sure that check passes before it merges a PR.

Additionally, I told bors to respect a `_ _ _` divider in the PR description as a way to tell it not to include anything below the divider in the commit message. 

This lets me easily override the commit title _and_ message for any PR before it's merged, making things easy on maintainers and on contributors! 🥳

orhun:
Great stuff!

PR #534, opened 2022-12-22
[Merged by Bors] - fix(ecs): abort program if a raw drop/clone implementation tries to panic over the C ABI

zicklag:
Panicking over the C ABI is undefined behavior, so it's better to be safe and abort the program instead.

Thanks to @LegionMammal978 on the Rust forum for finding this:

<https://users.rust-lang.org/t/review-for-my-small-1917-loc-rust-ecs/86326/2?u=zicklag>

### DIFF
--- a/crates/board-ecs/src/lib.rs
+++ b/crates/board-ecs/src/lib.rs
@@ -109,14 +109,36 @@ pub trait RawFns {
 
 impl<T: Clone> RawFns for T {
     unsafe extern "C" fn raw_drop(ptr: *mut u8) {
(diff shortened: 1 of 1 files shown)


zicklag:
bors merge

LegionMammal978:
Be careful: `eprintln!()` will unwind if it encounters an error while writing to `io::stderr()`. You may want to write `let _ = write!(io::stderr(), ...)` to avoid that.

zicklag:
Big thanks again, I'll do that.

PR #535, opened 2022-12-22
[Merged by Bors] - feat: add crab critters

ThousandthStar:
This implements the crabs environment feature. It uses the same logic as the code from when the game was using `macroquad`.


orhun:
🦀🦀🦀

src/map/elements/crab.rs
```diff
@@ -0,0 +1,204 @@
+use super::*;
+use rand::Rng;
+
+pub struct CrabDecorationPlugin;
+
+impl Plugin for CrabDecorationPlugin {
+    fn build(&self, app: &mut App) {
+        app.add_rollback_system(RollbackStage::PreUpdate, hydrate_crab)
+            .add_rollback_system(RollbackStage::Update, update_crabs);
```
zicklag, line 9: Here we also need to register the `CrabCritter` component with rollback plugin so it can be snapshot/restored in network games.

```suggestion
            .add_rollback_system(RollbackStage::Update, update_crabs)
            .extend_rollback_plugin(|plugin| {
                plugin
                    .register_rollback_type::<CrabCritter>()
            });
```

You will also need to derive `Reflect` on `CrabCritter` for that to work.

src/map/elements/crab.rs
```diff
@@ -0,0 +1,204 @@
...
+    mut crab_query: Query<(
+        &mut CrabCritter,
+        &mut AnimatedSprite,
+        &Transform,
+        &mut KinematicBody,
+    )>,
+    scary_things_query: Query<&Transform, (With<PlayerIdx>, Without<CrabCritter>)>,
+    time: Res<Time>,
+) {
+    let mut rng_bool = rand::thread_rng();
```
zicklag, line 114: For the random generation, try using the `GlobalRng` resource from the `crate::random` module.

That should be faster, and make sure that different computers in a network game generate the same random numbers.

zicklag (changes requested):
Thanks for working on this!

I've got a couple minor things to change, but this is looking good.

src/map/elements/crab.rs
ThousandthStar: I can't derive `Reflect` on `CrabCritter`, since `Reflect` can't be derived on `CrabState`. This is because `CrabState` is an enum. Should I represent the `crab_state` as a byte in the `CrabCritter` struct? @zicklag

zicklag: Try this to derive reflect on the enum:

```rust
#[derive(Reflect, Clone, Serialize, Deserialize)]
#[reflect_value(Serialize, Deserialize)]
pub enum Example {
    ExampleVariant,
}
```

zicklag (approved):
Looks good to me!

bors merge

PR #536, opened 2022-12-22
[Merged by Bors] - fix: make animation end frame index inclusive

zicklag:
The animation system was looping through the frames, not including the end frame in the animation, which is unintuitive.

This makes the animation system include the final frame.


zicklag:
bors merge

PR #541, opened 2022-12-30
[Merged by Bors] - feat: migrate game core to bones_ecs

zicklag:
This is a massive refactor that migrates to using `bones_ecs` for the core
game loop.

This splits the game into `jumpy` and `jumpy_core` with Jumpy being
bevy-based and responsible for running `jumpy_core` using `bones_lib`.
_ _ _

Work-in-progress migration to the the bones ECS for the core game loop.

PR #551, opened 2023-01-08
feat: upgrade to bevy 0.9

lecoqjacob:


zicklag:
This is finished in #541 but big thanks for your effort! ♥️

PR #558, opened 2023-01-16
[Merged by Bors] - feat: improve camera controller with camera_box follow technique.

zicklag:
This takes the player camera_box follow technique from the
pre-Bevy Jumpy version and combines it with the camera
lerp-ing technique that was added in the Bevy version.

This greatly improves the camera controller and avoids
excessive camera movement during play.


zicklag:
bors merge

PR #563, opened 2023-01-16
refactor: refactor to make sword implementation simpler.

zicklag:
- Update `bones_lib`.
- Create a system param for getting player inventories more conveniently.
- Use the new bones `Commands` queue to avoid an extra system.


zicklag:
bors merge

PR #567, opened 2023-01-17
[Merged by Bors] - feat: migrate crab to jumpy core.

lecoqjacob:
Migrates the crab from before the bones_ecs merge:

- Add back crab entity
- Add random back to jumpy core
- Fixes crab being stuck in fleeing state
_ _ _
Something I noticed while play testing is sometimes the crabs goes into a corner and doesn't move again. It does not happen every play through so there must be an edge case somewhere.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2504,6 +2504,7 @@ dependencies = [
  "hex",
  "serde",
  "tracing",
+ "turborand",
  "type_ulid",
 ]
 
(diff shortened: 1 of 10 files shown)


zicklag:
Awesome, thanks! 🎉 I'll test this out as soon as I get the chance!

zicklag:
bors merge

zicklag:
bors cancel

zicklag:
bors merge

PR #568, opened 2023-01-17
[Merged by Bors] - feat: add kick_bomb to jumpy core.

lecoqjacob:
Migrates the kick_bomb from before the bones_ecs merge.
_ _ _

- Add back kick_bomb

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,80 +1,67 @@
 [package]
-name = "jumpy"
-version = "0.5.1"
-description = "A tactical 2D shooter"
-authors = ["The Fish Fight Game & Spicy Lobster Developers"]
-license = "MIT OR Apache-2.0"
-edition = "2021"
+authors     = ["The Fish Fight Game & Spicy Lobster Developers"]
 default-run = "jumpy"
+description = "A tactical 2D shooter"
+edition     = "2021"
+license     = "MIT OR Apache-2.0"
+name        = "jumpy"
+version     = "0.5.1"
 
 [workspace]
-members = [
-    ".",
-    "core",
-]
+members = [".", "core"]
 
 [features]
 default = ["render"]
-render = [
-    "bevy/x11",
-    "bevy/png",
-    "bevy/filesystem_watcher",
-    "bevy/bevy_gilrs",
-]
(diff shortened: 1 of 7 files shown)


lecoqjacob:
jumpy_adMqQ8XwXF

zicklag:
Sweet! I might not get reviewing tonight, but I'll check as soon as I can.

Also, quick question, at the end of your video the screen went solid colored. Did you not re-spawn or something? I've never seen that happen. You usually respawn.

lecoqjacob:
It seems i did not respawn. I can do a deeper dive to see why

core/src/camera.rs
```diff
@@ -150,6 +149,12 @@ fn camera_controller(
         middle_point.y = size.y / 2.0;
     }
 
+    // Reset camera_pos to middle_point if it's NaN
+    let camera_pos = &mut camera_shake.center;
+    if camera_pos.is_nan() {
+        *camera_pos = middle_point.extend(0.0);
```
lecoqjacob, line 155: This solves the strange camera  issue with a single player dying. It would set the camera_pos to have an `x` of NaN causing all other math operations to be NaN. Resetting it to middle point for this edge case

lecoqjacob: New Camera:


https://user-images.githubusercontent.com/9278174/213054621-5cde980c-46a9-4718-888a-245feb79047a.mp4

zicklag: Ah, that makes sense. There must be a division by zero somewhere in the code when all the players are dead at the same time. I haven't played it one player since I update the camera controller so that's why I hadn't seen it before.

I'm glad it's not something that only happens on some computers because that would not be good!

Good job finding it. Do you mind opening a separate PR with just the camera fix?

lecoqjacob: https://github.com/fishfolk/jumpy/pull/569

lecoqjacob:
Kick bombs can now be kicked:

https://user-images.githubusercontent.com/9278174/213263811-6c401cfa-f694-4e3a-9202-0908845d456e.mp4

zicklag (approved):
Looks good! 👍

bors merge

zicklag:
bors retry

PR #571, opened 2023-01-18
[Merged by Bors] - feat: improve camera behavior when all players die.

zicklag:
Previously the camera position would get reset when all players
died and it would snap back into position when they re-spawned.
Now the camera stays in position until the players re-spawn and it
lerps to the new location smoothly.


zicklag:
Whoops forgot to merge this. :)

bors merge

PR #588, opened 2023-01-28
fix: fix sprite flip in attachments.

zicklag:


lecoqjacob:
@zicklag Should be fixed from my musket merge  https://github.com/fishfolk/jumpy/blob/b681945cd1e293c3a94f41ccb56cfdb78f84cbab/core/src/attachment.rs#L61

PR #591, opened 2023-01-30
[Merged by Bors] - refactor: move bullet meta to it's own asset type.

zicklag:


zicklag:
Hey @lecoqjacob I just wanted to get your review on this PR, because it's a refactor of the bullet metadata, and because I think it might effect our "best practices" for adding things in the future.

I think we'll actually want to do essentially the same thing for explosions, so that we can have explosion and bullet modules/systems/meta.

This should help us to be able to expand the parameters accepted in the bullet/explosion configuration, and maybe make it easier to add new gun types without having to write more code, similar to the old jumpy.

The idea is that we make the `BulletMeta` separate from `ElementMeta`, which I think is an important separation, because elements are supposed to represent things that may be placed on the map in the editor:

```rust
#[derive(BonesBevyAsset, Deserialize, Clone, Debug, Default, TypeUlid)]
#[ulid = "01GR1W2B3S7DM5QEY07RSJH2G0"]
#[asset_id = "bullet"]
#[serde(deny_unknown_fields)]
pub struct BulletMeta {
    #[serde(default)]
    pub body_offset: Vec2,
    // ...
}
```

Note that by deriving `BonesBevyAsset` instead of `BonesBevyAssetLoad` it get's it's own asset extension `.bullet.yaml`.

I also moved the bullet module out of the `elements` module, and was able to simplify it's [`hydrate`](https://github.com/zicklag/jumpy/blob/refactor/bullet-meta/core/src/bullet.rs#L30) function quite a bit, because we no longer have to treat it like an element or item, we just need to add the physics collider to it.

Anyway, I just at least wanted to make you aware of the change. :)

core/src/bullet.rs
```diff
@@ -12,105 +12,93 @@ pub fn install(session: &mut GameSession) {
...
+            entity,
+            Collider {
+                width: body_size.x,
+                height: body_size.y,
+                pos: transform.translation.truncate() + *body_offset,
+                ..default()
+            },
+        );
+
+        lifetimes.insert(entity, Lifetime::new(1.0));
```
lecoqjacob, line 78: should be default 1.0?

lecoqjacob (commented):
Very much like the change. It is simple, yet effective, plus slims down our elements quite a bit

core/src/bullet.rs
zicklag: 👍 I moved it to the `BulletMeta`.

zicklag:
Thanks for checking it out!

bors merge

PR #592, opened 2023-01-30
[Merged by Bors] - fix: prevent musket from buffering fire actions.

zicklag:
Previously, if you pressed the fire button quickly more than once,
the musket would fire twice, but slowly, when it should only fire
once.

This makes sure the musket doesn't buffer the fire attempt while it
is still in cooldown frames.


zicklag:
bors merge

PR #594, opened 2023-01-31
feat: reimplement ai

lecoqjacob:
@zicklag 

I adjusted the ruleset for the AI in `map.rs`. Seems like the more nodes we create, the weirder the AI starts reacting. If you play right now the AI moves around the whole map. Couple of things I have noticed.

I disabled the sideways jumps because they kept conflicting with each other causing a weird spin issue where the AI is switching from moving left and right repeatedly.

On certain parts of the map (very top) the ai gets stuck jumping up and down since the diagonal jumps are disabled

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2614,6 +2614,7 @@ dependencies = [
  "csscolorparser",
  "glam",
  "hex",
+ "petgraph",
  "serde",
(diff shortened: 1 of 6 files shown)


zicklag:
Closing these changes were worked into #665.

PR #598, opened 2023-02-02
[Merged by Bors] - build: reduce development opt-level.

zicklag:
This should decrease re-compile times for jumpy_core and jumpy.


zicklag:
bors merge

Issue #605, opened 2023-02-05
Upgrade to Rust v1.67.0


lecoqjacob:
Jumpy is locked in on 1.65.0, we need to upgrade and make sure it goes smoothly

PR #606, opened 2023-02-05
[Merged by Bors] - chore: upgrade rust to 1.67.0

lecoqjacob:
Closes: #605


Cargo.lock
```diff
@@ -2093,6 +2093,12 @@ dependencies = [
  "libc",
 ]
 
+[[package]]
+name = "hermit-abi"
```
zicklag, line 2097: I'm actually quite interested in why this package ended up in here. That seems strange.

zicklag: Ah, it looks like it's a dependency of `is_terminal` and `num_cpus`. In that case I don't think we need to worry about it.

zicklag:
bors merge

PR #607, opened 2023-02-06
[Merged by Bors] - feat: re-implement and juice crate.

ThousandthStar:
- Ports the crate to `bones_ecs`.
- Adds 2 breaking states (with sprites from the breaking animation;
  can be set in the config). 
- Adds 3 bounces before the crate breaks. A cooldown also makes it
  break if it hasn't broken from the states.
_ _ _

cc @zicklag


zicklag:
bors merge

zicklag (approved):
Thanks!

PR #616, opened 2023-02-08
feat: migrate collision detection to Parry.

zicklag:


core/src/camera.rs
```diff
@@ -84,7 +84,7 @@ fn camera_controller(
         }
 
         // Get the player's body rect
-        let rect = body.collider_rect(transform.translation);
+        let rect = body.aabb(*transform);
```
lecoqjacob, line 87: for those that dont know what aabb is, is there a way for us to document this for new developers to better understand? Maybe just a small comment

zicklag:
I added doc comments to all the physics methods, and I also found some more structs, fields, and a constant that we weren't using anymore that I removed.

lecoqjacob (approved):

PR #617, opened 2023-02-08
[Merged by Bors] - feat: port mine item to jumpy core.

DRuppFv:
_ _ _
The way the character holds it is weird, but I don't think it's hard to fix.


zicklag:
Awesome, thanks. 🚀 That's the last item we need for the MVP!

I think that the way he holds it is fine enough. I don't know how _I'd_ hold that honestly. Looks difficult to carry. 😜

bors merge

PR #618, opened 2023-02-08
[Merged by Bors] - feat: drop crate item where the player is holding it.

DRuppFv:
Update to drop crate item exactly where the player is holding it.
Also change the speed of the item when dropped.


zicklag:
bors merge

PR #631, opened 2023-02-17
[Merged by Bors] - refactor!: replace manual frame counts with timer on musket.

lecoqjacob:
_ _ _
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=pIi5lAe)

This was just a test to see the new time resource. I plan to go back and fix all other items post merge


assets/elements/item/musket/musket.element.yaml
```diff
@@ -4,7 +4,7 @@ builtin: !Musket
   atlas: ./musket.atlas.yaml
 
   max_ammo: 4
-  cooldown_frames: 50
+  cooldown_in_ms: 600
```
zicklag, line 7: Could you update this to just be `cooldown` and then use [`humantime_serde`](https://docs.rs/humantime-serde/latest/humantime_serde/) to deserialize it to a duration?

That way we can specify the cooldown as `600ms`.

core/src/elements/musket.rs
```diff
@@ -130,7 +132,8 @@ fn update(
             unreachable!();
         };
 
-        musket.cooldown_frame += 1;
+        println!("musket ammo: {:?}", musket.cooldown);
```
zicklag, line 135: Left over println.

core/src/session.rs
```diff
@@ -110,7 +110,7 @@ impl GameSession {
         let time_resource = self.world.resource::<Time>();
         time_resource
             .borrow_mut()
-            .advance_exact(std::time::Duration::from_secs_f32(crate::FPS));
+            .advance_exact(std::time::Duration::from_secs_f32(1.0 / crate::FPS));
```
zicklag, line 113: Good catch!

zicklag (changes requested):
Good idea!

Just got a couple small changes.

lecoqjacob:
@zicklag updated to use the new serde method. I like it alot!

zicklag (approved):
Love it!

On a sidenote, I just realized today that moving _all_ of our timing related code to use timers like this could be important, because it would allow us to run at a lower frames frames-per-second update during network games if necessary without changing the physics speed.

So this is definitely something we want to do for everything.

bors merge

zicklag:
bors merge

Issue #623, opened 2023-02-14
Level 4 Freezes


zicklag:
Probably due to the recent physics changes, level 4 freezes when loaded. I'm unsure as to the exact cause so far.

PR #632, opened 2023-02-17
[Merged by Bors] - fix: fix Level 4 freezing when loading.

iwek7:
The issue is that stomp boots were spawning in the wall
which made physics engine go crazy. 

Closes: https://github.com/fishfolk/jumpy/issues/623
_ _ _
In future some collision check should be added probably and if item is configured to spawn in the wall we could log some error and throw exception or something to be more explicit about what is wrong.


zicklag (approved):
Great, thanks!

There used to be a check that would make sure that the stuck in wall check wouldn't infinitely loop, but it must have gotten lost in the physics rework. I'll re-add that so this doesn't happen again.

bors merge

PR #637, opened 2023-02-17
[Merged by Bors] - fix: prevent freeze when shoving bodies out of tiles.

zicklag:
There was a scenario where a physics body could
cause an infinite loop if it ended up perfectly
aligned with the side of a tile. This prevents that
situation and panics if it ever happens again somehow.


zicklag:
bors merge

zicklag:
bors retry

Issue #635, opened 2023-02-17
Dropping the Sword In Place Kills You


zicklag:
In my attempt to fix how easily the sword kills you I apparently broke the scenario where you drop the sword in place.

We can probably fix this either by adding an arm delay again, or by increasing the kill velocity threshold.

PR #638, opened 2023-02-17
[Merged by Bors] - fix: increase killing velocity so that sword does not kill player when dropped in place

iwek7:
Tweaks sword drop position and throw velocity which
prevents you from killing yourself when dropping the
sword.

fixes: #635
_ _ _
I tested some sensible scenarios and it seems that throwing sword still kills enemy as expected. Value can be adjusted further down the road.

### DIFF
--- a/assets/elements/item/sword/sword.element.yaml
+++ b/assets/elements/item/sword/sword.element.yaml
@@ -8,9 +8,9 @@ builtin: !Sword
   grab_offset: [18, 26]
   body_size: [50, 8]
   # The minimum speed the sword must be moving to kill somebody
-  killing_speed: 7
+  killing_speed: 7.0
   angular_velocity: -0.04
   can_rotate: true
   bounciness: 0.32
-  throw_velocity: [1, 4]
+  throw_velocity: [1.5, 6]
   cooldown_frames: 22
--- a/core/src/elements/sword.rs
+++ b/core/src/elements/sword.rs
@@ -314,7 +314,7 @@ fn update(
 
             let transform = transforms.get_mut(entity).unwrap();
             transform.translation =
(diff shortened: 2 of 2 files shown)


zicklag:
Thank you!

I found a scenario where dropping the sword still kills you:

[drop-dead.webm](https://user-images.githubusercontent.com/25393315/219800393-8785efd8-2849-4251-ac0d-142435864600.webm)

It seems to happen because when you drop the sword of the side of the block, it falls further and therefore accumulates more speed.

**But** I also just realized that when we drop the sword, it's dropping from above our head, so if we dropped it at waist level it shouldn't be able to accumulate enough speed before stopping colliding with us. And it'd be better for it to drop from lower down for looks anyway.

Could you try adjusting the drop position of the sword and see if that fixes it?

iwek7:
Okey I will do this, with current setup I was forced to move velocity up to 8.6 which is a bit too high imo

iwek7:
So actually when sword was thrown from above head it had this nice side effect that when you run and throw sword it sort of goes up over you and does not kill you instantly, For this to work from waist level I had to tweak throw velocity. Now it has similar feel to previous throwing and also does not kill you.
Also killing speed velocity change is not needed in this scenario, sorry for multiple pushes but I just keep turning this params and I think some config works better and I keep changing it. Now this is final solution.

zicklag:
> Also killing speed velocity change is not needed in this scenario, sorry for multiple pushes but I just keep tweaking those params and I keep thinking some config works better and I keep changing it. Now this is final solution.

No problem at all! It's a tricky little problem.

This is much better than before, thanks!

bors merge

zicklag:
bors cancel

zicklag:
bors merge

PR #641, opened 2023-02-22
[Merged by Bors] - feat: add slippery seaweed

Zac8668:


core/src/player/state/states/incapacitaded.rs
```diff
@@ -0,0 +1,79 @@
...
+pub const ID: Key = key!("core::incapacitaded");
+const SLOWING_SPEED: f32 = 0.3;
+
+pub fn player_state_transition(
+    entities: Res<Entities>,
+    slippery_seaweeds: CompMut<SlipperySeaweed>,
+    collision_world: CollisionWorld,
+    mut player_states: CompMut<PlayerState>,
+) {
+    for (seaweed_ent, _) in entities.iter_with(&slippery_seaweeds) {
```
zicklag, line 13: Could you move this code to the `slippery_seaweed.rs` module?

That way if we add new items or things that can send the player into the incapacitated state, they can each be in their own module, and we don't have to keep making this system bigger to account for each new item.

core/src/player/state.rs
```diff
@@ -47,6 +47,7 @@ pub fn install(session: &mut GameSession) {
     add_state_module!(session, midair);
     add_state_module!(session, walk);
     add_state_module!(session, dead);
+    add_state_module!(session, incapacitaded);
```
zicklag, line 50: _( Read comment below about moving the state transition function first. )_

Since we don't need a `player_state_transition` system in the `incapacitated` state, don't use the module here and just register the state handler system:

```suggestion
    session
        .stages
        .add_system_to_stage(CoreStage::PreUpdate, incapacitated::handle_player_state);
```

zicklag (changes requested):
Looking good, thanks!

Just a couple of things to tweak.

core/src/player/state.rs
```diff
@@ -47,6 +47,11 @@ pub fn install(session: &mut GameSession) {
     add_state_module!(session, midair);
     add_state_module!(session, walk);
     add_state_module!(session, dead);
+
+    session
+        .stages
+        .add_system_to_stage(CoreStage::PreUpdate, incapacitated::handle_player_state)
+        .add_system_to_stage(PlayerStateStage, slippery_seaweed::update);
```
zicklag, line 54: Tiniest little change: it'd be good to move the registration of the slippery seaweed system to it's own `install` function.

That should be the last thing!

```suggestion
        .add_system_to_stage(CoreStage::PreUpdate, incapacitated::handle_player_state);
```

core/src/elements/slippery_seaweed.rs
```diff
@@ -0,0 +1,82 @@
+use crate::prelude::*;
+
+pub fn install(session: &mut GameSession) {
+    session
+        .stages
+        .add_system_to_stage(CoreStage::PreUpdate, hydrate);
```
zicklag, line 6:
```suggestion
        .add_system_to_stage(CoreStage::PreUpdate, hydrate)
        .add_system_to_stage(PlayerStateStage, update);
```

zicklag (changes requested):

core/src/player/state.rs
Zac8668: I tried to do this first actually, as it makes more sense, but I get this error:
`thread 'main' panicked at 'Stage with label `PlayerStateStage` ( 01GP72HD5B1S7VDNE9B2SAGWVX ) doesn't exist.'`
Looks like this install function runs before the PlayerStateStage is created, not sure how to get around this.

Zac8668: It worked now :)
(not sure how often you check discord, but I've asked you something on the jumpy channel, as you're awnsering fast here, just letting you know)

zicklag (approved):
Perfect, thanks!

bors merge

PR #643, opened 2023-02-24
[Merged by Bors] - fix: fix world inspector warnings at startup.

zicklag:
We don't enable `bevy_pbr` so there were world
inspector warnings about those types not being
registered.

This avoids running the code that looks for the
`bevy_pbr` types.


zicklag:
bors merge

PR #645, opened 2023-02-24
fix: reset item state on out-of-bounds re-spawn.

zicklag:


zicklag:
Closing so @TekhnaeRaav can open the PR from his account.

Issue #628, opened 2023-02-16
Tweak Item Throwing Velocities


zicklag:
From @Hyeve-jrs ( I think, `Hyeve#6138` technically ) on Discord:

> and throwing is.. not actually based on player momentum in duck game, though it usually looks like it
throw speeds are actually determined by the inputs you're holding - you will always do a fast throw if you're holding left or right, regardless of how fast you're actually moving, and similarly always do a slow throw if not holding left or right
and straight up / down if you're holding up or down
(technically there's also a case for holding left/right and up, which will do a fast throw with a higher arc)

Right now our item throwing is based on the speed of movement, but it should be done as described above.

zicklag:
My bro @tekhnaeraav is going to try to take this one.

PR #646, opened 2023-02-24
feat: add new item throw behaviors.

zicklag:
Add differing throw types and add their functionality to the
current item selection.

Convert the PlayerEvent enum into a struct with systems for use in
Commands. This replaces the handle_player_events system because of
timing issues.

Closes: #628


zicklag:
Closing so @TekhnaeRaav can open the PR from his account.

PR #648, opened 2023-02-24
[Merged by Bors] - feat: animate crabs when walking.

TekhnaeRaav:
The crabs will now animate at the specified fps while walking and twice
that while fleeing and also freeze when paused.


zicklag:
bors merge

PR #650, opened 2023-02-25
[Merged by Bors] - feat: re-add camera shake to items.

DRuppFv:
_ _ _
After this, I'll make a major PR, I just need this to be merged first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -942,7 +942,7 @@ checksum = "0d8c1fef690941d3e7788d328517591fecc684c084084702d6ff1641e993699a"
 [[package]]
 name = "bones_asset"
 version = "0.1.0"
-source = "git+https://github.com/fishfolk/bones#7e00c6e7b6300054ffeeebd186b5adf96b8aa10b"
+source = "git+https://github.com/fishfolk/bones#473af7fcea22f1c76a84dd3f243fc5c282b142a5"
 dependencies = [
  "bevy_asset",
  "bones_bevy_utils",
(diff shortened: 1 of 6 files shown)


zicklag (approved):
Thanks! This looks good. I merged https://github.com/fishfolk/bones/pull/104 into bones first, and I'm pushing up my tweaks to the camera shake parameters.

Also I'm removing the camera shake on the crate for now. During testing, me and @TekhnaeRaav felt that the crate doesn't feel right shaking the ground like an explosion would.

We can tweak the prarameters more, too, if we get feedback saying it doesn't feel right.

bors merge

DRuppFv:
> Thanks! This looks good. I merged [fishfolk/bones#104](https://github.com/fishfolk/bones/pull/104) into bones first, and I'm pushing up my tweaks to the camera shake parameters.
> 
> Also I'm removing the camera shake on the crate for now. During testing, me and @TekhnaeRaav felt that the crate doesn't feel right shaking the ground like an explosion would.
> 
> We can tweak the prarameters more, too, if we get feedback saying it doesn't feel right.
> 
> bors merge

Ok. I've been thinking of adding camera shake on the musket (only to the shooter), but this answer made me wonder if you would use this. What do you think?

zicklag:
Yeah, I'm thinking we probably don't want to add camera shake to the musket, but I can only speak for me and @TekhnaeRaav. It seems best to me to keep it to explosions for now.

@erlend-sh or @legendiguess do you have any thoughts?

erlend-sh:
Yeah during initial testing we were fine with applying the shale more liberally, but it should really just be used for explosions and other high impact weapons.

PR #652, opened 2023-02-26
[Merged by Bors] - chore: increase default optimization level during development.

zicklag:
It seems that we need at least a little optimization
for good performance during development.

This can still be turned down by developers if they
need maximum compile speed.


zicklag:
bors merge

PR #655, opened 2023-02-27
[Merged by Bors] - fix: improve fixed-timestep behavior and fix physics performance issues.

zicklag:
- updates the fixed-timestep behavior to play the game
  in slow-motion instead of completely freezing the game
  when it can't keep up.
- fix a rapier collision detection performance issue causing
  problems while map editing.
- Fix and rename lev03 to Level 13, moving it to the default
  maps.


zicklag:
bors merge

PR #657, opened 2023-02-27
[Merged by Bors] - fix: fix ship_decorations tilemap.

zicklag:
The atlas size was incorrectly set in the YAML,
causing it to behave strangely.


zicklag:
bors merge

Issue #633, opened 2023-02-17
Add Padding Pixels to Sproinger Sprite Sheet


zicklag:
Due to #446 the sproinger's sproinging animation has some bleed.

An easy temporary fix would just be to add 1 pixel of transparent padding around the sprite sheet.

zicklag:
Closed in #664.

PR #664, opened 2023-03-01
fix: add padding pixels to atlases to prevent bleeding.

lvllvl:
I was attempting to resolve issue #633. I increased the size of the the following yamls: 
* springer.atlas.yaml 
* crab.atlas.yaml 

I apologize if this isn't *exactly* the change that you're looking for ( Or if I misunderstood the direction of the issue altogether ). 

Please let me know if I should make any additional changes.


assets/elements/environment/crab/crab.atlas.yaml
```diff
@@ -1,4 +1,4 @@
 image: ./crab.png
-tile_size: [17, 12]
+tile_size: [17, 13]
```
zicklag, line 2: This should actually be:

```suggestion
tile_size: [18, 12]
```

zicklag (commented):
Thanks for trying this out!

You also need to update the sprite sheet to add an extra row of blank pixels to the top of every tile in the sprite. For example, with the sproinger:

image

The red rectangles show where we need to add an extra row of blank pixels for padding.

This will make the sprite 3 pixels taller than before and make sure that we've got blank space between each tile to prevent the bleeding effect.

For the crabs we actually need to make the tiles _wider_ not taller.

image

We need to add two columns of blank sprites like showin in the image above. So the crab tile size should be made one pixel wider, not one pixel taller.

assets/elements/environment/crab/crab.atlas.yaml
lvllvl: Hi @zicklag, thanks for the feedback. 

Ok so I gave it another shot. 

I updated the following: 

* crab.atlas.yaml to match your suggestions
* crab.png + sproinger.png to match your requirements as well. 

Hopefully this works, but again if changes need to be made please let me know.

zicklag:
Thanks! I've been a little busy, but I should have a chance to look at this tomorrow.

zicklag:
Checked it out!

The crabs are perfect. 👍

The sproinger needs a couple tweaks. One of them was my fault. I thought the sproinger didn't need any left-and-right padding, but apparently it had some left-and-right bleed, too.

So we need to increase the tile size width to 33px. We also need to make sure the sprite size matches the new tile size.

So since our new tile size is 33x21, we need to make sure that the width of the sprite is 66 pixels because our sprite is two tiles wide, and our height is 63 pixels, because our sprite is 3 tiles high.

Then we also need to make sure that we move the sproinger into position in each tile, so that there is a gap on each side of the sproinger, and that the bottom of the sproinger is lined up with the other tiles.

I recommend [libresprite](https://libresprite.github.io/#!/downloads) for sprite editing, but you can use anything.

I made a small video showing what changes to make to the sproinger.

**Note:** My calculator prompt pops up in the middle of the video, but all I'm doing with it is calculating what the sprite size should be.

[jumpy-sproinger-fix-2023-03-07_16.20.14.webm](https://user-images.githubusercontent.com/25393315/223567389-0b07421b-e14b-4b75-ba86-3ea7fbcb9202.webm)

zicklag:
I pushed e858922deac71776b705746ebc6d6f38dcbe38e8 which updated the tile size in the YAML, and just slightly shifted the positions of the sproingers until it looked right when running the game. It's ready to go now.

Thanks for helping out! I know I ended up walking through how to do it and such, but don't let that discourage you at all, and we definitely appreciate your time spent to help our project. Doing things is one of the best ways to learn, so don't be afraid to try new stuff!

Welcome to the Jumpy contributors. 😃 :man_astronaut:

lvllvl:
awesome! thanks again for all your help.

PR #671, opened 2023-03-02
[Merged by Bors] - fix: revert attempted asset loading fix and implement web-compatible fix.

zicklag:
Unfortunately the web platform can't read out all the assets in a folder
so it will always fail to load on the browser.

I switch to an alternative strategy to fix the starutp crash by making
sure only the assets we are going to use are loaded, which was what we
used to do, but I had forgotten to check for other egui textures as I
added them to the loading process.


zicklag (commented):
bors merge

PR #673, opened 2023-03-03
feat: add slippery tile.

Zac8668:


zicklag (commented):
This works great! I pushed a minor update that adds the slippery and slippery seaweed to the editor.

---

I'm fine merging this for now if you don't want to do this yet, but I was thinking that it'd be good to add a per-body friction option to the physics instead of the global friction parameter we have today.

Then, instead of adding the slipperyness logic to the player controller, we could cause the slippery elements to cause anything colliding with them to have a decreased friction, that would be re-set to normal when they stop colliding.

That would make everything slip on it, including items.

But I'm fine merging this now and just opening an issue for future work if you'd rather do that.

core/src/physics.rs
```diff
@@ -56,6 +56,12 @@ pub struct KinematicBody {
     pub angular_velocity: f32,
     pub gravity: f32,
     pub bounciness: f32,
+    /// Sets a 1 frame override for the body friction. It will be re-set to `None` every frame so if
+    /// you wish to apply a continuous friction change, you must re-set it every frame.
+    ///
+    /// This is useful for things like slippery blocks or other things that want to modify a body's
+    /// friction while it is on the block.
+    pub frame_friction_override: Option<f32>,
```
zicklag, line 64: Awesome, thanks for the update!

I pushed a slight tweak that renamed the `friction` field and changed it to an `Option` to try and be more clear about how it's used.

I just wanted to make sure that looks good to you before merging.

Zac8668: Looks good to me 👍

PR #693, opened 2023-03-09
feat: add invincibility on spawn

lecoqjacob:
@zicklag Here is the invincibility! I went ahead and added a config to the core meta, but let me know if i should move it


lecoqjacob:
This was testing with 5 second invincibility

https://user-images.githubusercontent.com/9278174/224120042-4dd29a09-19db-499f-b4b4-e560099d8f36.mp4

something.yaml
```diff
@@ -0,0 +1,3 @@
+|
```
zicklag, line 1: Looks like this got left over. 🙂

zicklag (changes requested):
Looks like there's an accidentally included file, but otherwise this looks good!

Thanks!

something.yaml
lecoqjacob: Whoops! was testing what a duration was serialized to

zicklag (approved):

Issue #697, opened 2023-03-09
Reworking levels


SirGiraffey:
Reworking and redesigning all levels to use the newer tilemaps, critters, weapons etc. using the new editor.

Here's the first 3!

[Level 1.txt](https://github.com/fishfolk/jumpy/files/10936391/Level.1.txt)
[Level 2.txt](https://github.com/fishfolk/jumpy/files/10936392/Level.2.txt)
[Level 3.txt](https://github.com/fishfolk/jumpy/files/10936393/Level.3.txt)

I'm attaching them as txt files, let me know if that suits you

zicklag:
Awesome! That'll work fine for me. 👍

zicklag:
@SirGiraffey I was able to re-order the layers in Level 3 so you get this nice foreground action where some times can be on top of the players/items/critters:

![image](https://user-images.githubusercontent.com/25393315/224177686-56049b80-a153-46b5-a415-acdf08aacf7a.png)

Just wanted to make sure you wanted it like that.

Also, I kind of liked the old level 2. Do you think we could take your new level 2 and add it as a new map so we can keep the old one, too?

SirGiraffey:
Hey, I remade a few next levels. I love the foreground effect you did with the nets.
And I'll remake the 2. level again then, without changing it too much :)

[Level 4.txt](https://github.com/fishfolk/jumpy/files/10962622/Level.4.txt)
[Level 5.txt](https://github.com/fishfolk/jumpy/files/10962623/Level.5.txt)
[Level 6.txt](https://github.com/fishfolk/jumpy/files/10962624/Level.6.txt)
[Level 7.txt](https://github.com/fishfolk/jumpy/files/10962619/Level.7.txt)
[Level 8.txt](https://github.com/fishfolk/jumpy/files/10962620/Level.8.txt)
[Level 9.txt](https://github.com/fishfolk/jumpy/files/10962621/Level.9.txt)

zicklag:
Oh sweet! I'm just getting ready to sit down with my bros and do some testing, so we'll try out your new levels, too!

SirGiraffey:
Sending 2 more levels!

[Level 10.txt](https://github.com/fishfolk/jumpy/files/11012399/Level.10.txt)
[Level 11.txt](https://github.com/fishfolk/jumpy/files/11012401/Level.11.txt)

By the way, do you have any feedback from testing the other levels?

zicklag:
[...]

PR #700, opened 2023-03-09
feat: incorporate @SirGiraffey's updated levels 1 and 3 and add level 14.

zicklag:


zicklag (commented):
@SirGiraffey I left level 2 unchanged and added your level2.txt as level 14 if you think that makes sense.

zicklag:
Just pushed the rest of the maps from #697. I'm going to merge what we have, and we can add more as we go.

Issue #712, opened 2023-03-13
Don't Let Second Player Controls Control AI Players


zicklag:
Right now if you add an AI player for player 2 you can still control them with the keyboard using the player 2 controls. We should make it impossible to control AI players.

lvllvl:
1. do you think this is simply a matter of disabling keys? If so, which keys would be considered player 1? Like is there a preference for which keys to disable?
2. would I need to make the changes in the `player_ai_system` function ( core > source > player.rs)

zicklag:
This should actually be super simple. You should just have to do a check here for whether or not the player is an AI, and just `continue` the loop if they are an AI:

https://github.com/fishfolk/jumpy/blob/cd0fb59e3ddf8b0f5d22bc292fcfa53f02e10759/src/session.rs#L149-L150

Something like:

```rust
if inputs.players[player_idx.0].is_ai { continue; }
```

PR #721, opened 2023-03-14
changed session.rs so ai player cannot be controled.

lvllvl:
Updates based on issue #712. 

1. Tested both online demo and launched game from my terminal to see how I could control ai player. Generally it was through the `jump` key, and sometimes the `left` and `right` keys, but to a lesser extent. 
2. Updated session.rs `line 151` to stop receiving user input. 

Does this project require tests being added to make sure the code is correct / stable?

If so I can try and add that to my PR. Please lmk if any additional changes are required. 

Thanks!

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -147,6 +147,9 @@ fn update_input(
         inputs.players[0].editor_input = editor_input.take();
 
         for (player_idx, action_state) in &player_input_collectors {
+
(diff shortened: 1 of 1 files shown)


zicklag (approved):
Looks good thanks!

Right now we don't have any testing infrastructure setup so this should be good.

PR #725, opened 2023-03-16
fix: prevent bullet from killing shooter.

### LINKED ISSUE #715: Don't Let The Musket Bullet Kill the Player That Shot It
This can happen if you turn around very quickly.

We can fix this by setting the owner of the damage region for the bullet to the player that shot the gun.

iwek7:
This fixes #715. Issue suggested that what should be done is adding owner to damage region (as is done in case of sword), however it seems that bullet does not use those regions so I simply add entity marker to bullet object and filter out all collisions with this entity.

### DIFF
--- a/core/src/elements/musket.rs
+++ b/core/src/elements/musket.rs
@@ -221,6 +221,7 @@ fn update(
                             bullets.insert(
                                 ent,
                                 Bullet {
+                                    owner: player,
                                     direction: if player_flip_x { -1.0 } else { 1.0 },
                                 },
                             );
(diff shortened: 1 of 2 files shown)


core/src/elements/musket.rs
```diff
@@ -221,6 +221,7 @@ fn update(
                             bullets.insert(
                                 ent,
                                 Bullet {
+                                    owner: player.clone(),
```
zicklag, line 224: Can you remove this `clone()`.

zicklag (commented):
Awesome, thanks! Just one clippy lint to fix.

And please run `cargo fmt`, too.

Issue #717, opened 2023-03-14
Make The Camera Left/Right Padding Larger


zicklag:
On certain really wide maps, it becomes harder to jump to the left and right of the screen, where the camera doesn't show. We need to make the camera zoom out a little more to give the players more breathing room when playing near the edge of a wide map.

---

We should just need to increase these parameters:

https://github.com/fishfolk/jumpy/blob/cd0fb59e3ddf8b0f5d22bc292fcfa53f02e10759/assets/default.core.yaml#L6-L7

And then test a two player game on one of the super-wide maps.

zicklag:
Experiment to see what works best.

In particular, if you start level 6 with two players you can get into a situation where there's a platform to the right of the camera, but you can't really see it, so it makes it hard to jump up on to it:

![image](https://user-images.githubusercontent.com/25393315/225774989-e3b6daeb-0983-40b3-b17d-db7eb263edc9.png)

We want a little bit more space so that you can see what you're about to jump up onto.

kaminobolt:
I increased the padding to 300 so it is visible ahead while running

pull request #727

PR #726, opened 2023-03-17
fix: updated default.core.yaml to make camera padding larger

lvllvl:
Updated the camera padding in `default.core.yaml` to make view wider, especially on level 6. 
Screenshot 2023-03-16 at 8 01 46 PM

Fix for #717


zicklag:
I tested this PR and #727 and #727 seemed to work better, so I'm merging that one, but thank you still!

PR #727, opened 2023-03-17
fix: increased camera padding to 300px.

kaminobolt:
300 seems like a better option when moving faster

Fixes: #717


zicklag:
Looks good, thank you!

PR #728, opened 2023-03-17
feat: let the AI choose their enemies and chase them.

Plecra:
this replaces the old logic which had all AI chase player 1. It's a small improvement which means they'll attack each other sometimes, and player 1 doesn't always have to be running away


zicklag:
Oh, cool, thanks!

Looks good upon a quick check, and I'll test it out tonight or tomorrow. 👍

zicklag (approved):
Works great, thanks!

Issue #723, opened 2023-03-15
Grenades That Explode While You Hold Them Fly Out of Your Hand Instead of Despawning


zicklag:

lecoqjacob:
Ill take this

zicklag:
Awesome. Just a warning, @TekhnaeRaav tried to look at this at some point and thought it might have beena weird ordering thing with Bones, or maybe something like the fact that a component was reading stale data, along the lines of https://github.com/fishfolk/bones/issues/96 or https://github.com/fishfolk/bones/issues/82.

I haven't looked at it yet, so I'm not sure, but it's something to keep an eye out for.

PR #732, opened 2023-03-18
fix: don't drop killed items.

lecoqjacob:
Fixes: #723


core/src/elements/grenade.rs
```diff
@@ -209,20 +206,6 @@ fn update_lit_grenades(
             let player = inventory.player;
             let layers = player_layers.get_mut(player).unwrap();
             layers.fin_anim = *fin_anim;
-            let body = bodies.get_mut(entity).unwrap();
```
zicklag, line 212: Is this not needed for some reason?

lecoqjacob: I believe our attachment system already handles all of these params for us

zicklag (approved):
Awesome, thanks for doing the deep dive on this!

Issue #731, opened 2023-03-18
Throwing a Crate Up Should Be Able to Bonk Yourself If You Don't Move.


zicklag:
We might need to do something like an "arm delay" so that you can't hurt yourself with the crate, until a split second after it's thrown.

lvllvl:
I was looking around in the codebase, does this issue entail changing:

* `core > elements > crate_items.rs`
    + `fn update_thrown_crates`

Does `core>src> damage.rs` also get updated in this?

When a box hits the top of a player's head, does it cause damage? Or kill the player? Or does the box just break and respawn?

zicklag:
> changing... `fn update_thrown_crates`

Yeah, the key line is this one here:

https://github.com/fishfolk/jumpy/blob/b22603540980b702212fbfa1b3601cb5dfeb0532/core/src/elements/crate_item.rs#L218

Where it makes sure `e != thrown_crate.owner` is how it makes sure it doesn't damage the person who threw the crate. We need to make sure that the crate kills the player if it hits them out of mid-air, but doesn't kill them immediately when they throw it out of their hands.

We don't need to change `damage.rs`. When it hits the players head it kills theplayer, and the box breaks and re-spawns.

PR #733, opened 2023-03-20
fix: make box kill player when box hits the player on the head.

lvllvl:
When player_b throws a box directly up and the box collides with player_b's head, the box now kills player_b.

Fixes: #731


core/src/elements/crate_item.rs
```diff
@@ -215,7 +215,10 @@ fn update_thrown_crates(
 
         let colliding_with_players = collision_world
             .actor_collisions_filtered(entity, |e| {
-                e != thrown_crate.owner && players.contains(e) && invincibles.get(e).is_none()
+                e == thrown_crate.owner
```
zicklag, line 218: This will make it _only_ kill the player who threw the crate, which isn't what we want.

What we should probably do is:

- remove this `e == thrown_crate.owner` check so that it will detect collisions with all players.
- in the loop below `for player_entity in &colliding_with_players`, we do an extra check to see if `e == thrown_crate.owner`
- if it hits the crate owner, then only kill the owner if `thrown_crate.age > 0.25`
- that way the person who throws the crate is only killed if the crate has been in the air for more than a quarter second.

You may need to mess with that 0.25 value and test it in game to make sure that's a good amount of time that keeps you from killing yourself right when you throw the item.

lvllvl: ok cool, ty for the feedback. I'll start making these changes this afternoon.

lvllvl: Ok I tried to implement your feedback. 

In the `for player_entity in &colliding_with_players` loop, I tried to include `if-statements` that would account for both: 
* killing the owner under the appropriate circumstances 
* killing other entities under the appropriate circumstances 

0.25 seconds seems to be a decent amount of time. 

Please let me know if you'd like me to change anything else.

lecoqjacob:
Can we switch age from manually deriving time to the new Bones time resource? I believe musket and invincibility currently utilize this

core/src/elements/crate_item.rs
```diff
@@ -215,16 +215,25 @@ fn update_thrown_crates(
 
         let colliding_with_players = collision_world
             .actor_collisions_filtered(entity, |e| {
-                e != thrown_crate.owner && players.contains(e) && invincibles.get(e).is_none()
+                players.contains(e) && invincibles.get(e).is_none() && thrown_crate.was_colliding
```
zicklag, line 218:
```suggestion
                players.contains(e) && invincibles.get(e).is_none() && thrown_crate.was_colliding
```
```suggestion
                players.contains(e) && invincibles.get(e).is_none()
```

zicklag (changes requested):

core/src/elements/crate_item.rs
erlend-sh: > Should the player that throws the crate live and the other player die? Or should they both survive?

They should both die.

core/src/elements/crate_item.rs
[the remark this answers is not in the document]
zicklag: Erlend [said](https://github.com/fishfolk/jumpy/pull/733#discussion_r1143769396) that both of the players need to die if you throw a box while it is colliding with another player. That makes this just a little more complicated.

You'll need to get this behavior:

- When throwing a box away from yourself, it doesn't kill you.
- When throwing a box up, and it lands on your head, it kills you.
- When getting hit by a box in mid-air, you die.
- When throwing a box while another player is colliding with the box in your hand, the box should kill both you and the player you hit the box with.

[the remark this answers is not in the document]
lvllvl: ok cool, I'll try and implement this. I realized that I was testing only with player 2. When I tried this with player 1, you're right it doesn't work at all. 

Ok I'll work on this. Thanks all for giving feedback.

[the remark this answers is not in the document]
lvllvl: hi, 

So I added the following checks: 

- When throwing a box away from yourself, it doesn't kill you.
- When throwing a box up, and it lands on your head, it kills you.
- When getting hit by a box in mid-air, you die.

For the following requirement, I added a new function at the bottom of the file. 
- When throwing a box while another player is colliding with the box in your hand, the box should kill both you and the player you hit the box with.


I was also able to refactor the `thrown_crates` struct to match the implementation of bones time resource, like in `musket.rs`. 

Please let me know if I should change anything. And thanks for your help!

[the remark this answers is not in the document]
zicklag: Awesome, I'll check it out if I can tomorrow, thanks!

core/src/elements/crate_item.rs
```diff
@@ -226,9 +234,18 @@ fn update_thrown_crates(
...
+            &players,
+            &invincibles,
+            &mut commands,
+            transform,
+        );
 
         if !colliding_with_players.is_empty()
-            || thrown_crate.age >= *break_timeout
+            || kill_nearby_colliding
+            || thrown_crate.damage_delay.elapsed_secs() >= *break_timeout
```
zicklag, line 248: Ah, I guess we should have a second `break_timeout` timer here, too. So that we can do this:

```suggestion
            || thrown_crate.break_timeout.finished()
```

Then on line 130, we create the timer using the `break_timeout` duration from the asset metadata.

I think that's the last change.

zicklag (changes requested):
I found one more change, and then I think this'll be done!

Also, whenever you make some changes and are ready for me to look at it again, click this little rotate button. That will tell me that it's ready for a re-review.

image

core/src/elements/crate_item.rs
lvllvl: 1. Ok I made the break_timeout change. on lines 114-125, I added what I thought was asset metadata. If that's not what you meant, I can change it. 

2. Does this file need to be updated as well: `jumpy/assets/elements/item/crate/crate.element.yaml`, like break_timeout: 4s, instead of just 4?

3. Also updated return structure on the new `fn kill_all_colliding_..` because the crates were breaking too soon the way I had it initially.

core/src/elements/crate_item.rs
```diff
@@ -179,7 +206,7 @@ fn update_thrown_crates(
             continue;
         };
 
-        thrown_crate.age += 1.0 / crate::FPS;
+        thrown_crate.damage_delay.tick(time.delta());
```
zicklag, line 209:
```suggestion
        thrown_crate.damage_delay.tick(time.delta());
        thrown_crate.break_timeout.tick(time.delta());
```

core/src/elements/crate_item.rs
```diff
@@ -167,7 +194,7 @@ fn update_thrown_crates(
             breaking_anim_frames,
             breaking_atlas,
             breaking_anim_fps,
-            break_timeout,
+            // break_timeout, TODO: delete this
```
zicklag, line 197:
```suggestion
```

zicklag (changes requested):
> Does this file need to be updated as well: jumpy/assets/elements/item/crate/crate.element.yaml, like break_timeout: 4s, instead of just 4?

Yeah, that'd be good.

zicklag (approved):
This looks good, thanks!

PR #736, opened 2023-03-22
fix: fix musket reloading when dying holding it.

### LINKED ISSUE #719: Musket Doesn't Reload When Player Dies While Holding It

Zac8668:
Fixes: #719


zicklag (approved):
Thanks!

PR #742, opened 2023-03-26
refactor: use bones timer for kick bomb fuse and arm delay.

### LINKED ISSUE #741: Update kick_bomb to use Bones time resource instead of manually setting the time
Update kick_bomb to use Bones time resource instead of manually setting the time. 

Note: I know other elements need this as well. I thought about doing it all at once but I didn't want to assume so I've just been doing it one element at a time.

lvllvl:
Added bomb timer to kick_bomb.rs instead of manually setting the time. 

Fixes: #741

### DIFF
--- a/core/src/elements/kick_bomb.rs
+++ b/core/src/elements/kick_bomb.rs
@@ -209,7 +209,7 @@ fn update_lit_kick_bombs(
             unreachable!();
         };
 
-        kick_bomb.arm_delay.tick(time.delta());
+        kick_bomb.fuse_time.tick(time.delta());
 
         let mut should_explode = false;
         // If the item is being held
@@ -259,7 +259,7 @@ fn update_lit_kick_bombs(
             } else if !player_standing_left && player_sprite.flip_x {
                 body.velocity.x = -kick_velocity.x;
                 body.velocity.y = kick_velocity.y;
-            } else if kick_bomb.arm_delay.finished() || kick_bomb.fuse_time.finished() {
+            } else if kick_bomb.fuse_time.finished() {
                 should_explode = true;
             }
         }
--- a/core/src/elements/kick_bomb.rs
+++ b/core/src/elements/kick_bomb.rs
@@ -1,5 +1,4 @@
 use crate::prelude::*;
-use std::time::Duration;
 
 pub fn install(session: &mut GameSession) {
     session
@@ -147,14 +146,8 @@ fn update_idle_kick_bombs(
                     lit.insert(
                         entity,
                         LitKickBomb {
-                            arm_delay: Timer::new(
-                                arm_delay,
-                                TimerMode::Once,
-                            ),
-                            fuse_time: Timer::new(
-                                fuse_time,
-                                TimerMode::Once,
-                            ),
+                            arm_delay: Timer::new(arm_delay, TimerMode::Once),
+                            fuse_time: Timer::new(fuse_time, TimerMode::Once),
                         },
                     );
                 },
--- a/core/src/elements/kick_bomb.rs
+++ b/core/src/elements/kick_bomb.rs
@@ -1,4 +1,5 @@
 use crate::prelude::*;
+use std::time::Duration;
 
 pub fn install(session: &mut GameSession) {
     session
@@ -12,10 +13,10 @@ pub fn install(session: &mut GameSession) {
 #[ulid = "01GQ0ZWBNA8HZRXYKZXCT05CXT"]
 pub struct IdleKickBomb;
 
-#[derive(Clone, TypeUlid, Debug, Copy)]
+#[derive(Clone, TypeUlid, Debug)]
 #[ulid = "01GQ0ZWFYZSHJPESPY9QPSTARR"]
 pub struct LitKickBomb {
-    age: f32,
+    damage_delay: Timer,
 }
 
 fn hydrate(
@@ -137,7 +138,12 @@ fn update_idle_kick_bombs(
             commands.add(
                 move |mut idle: CompMut<IdleKickBomb>, mut lit: CompMut<LitKickBomb>| {
                     idle.remove(entity);
-                    lit.insert(entity, LitKickBomb { age: 0.0 });
+                    lit.insert(
+                        entity,
+                        LitKickBomb {
+                            damage_delay: Timer::new(Duration::from_secs_f32(8.0), TimerMode::Once),
+                        },
+                    );
                 },
             );
         }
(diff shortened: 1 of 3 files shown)


core/src/elements/kick_bomb.rs
```diff
@@ -137,7 +138,12 @@ fn update_idle_kick_bombs(
             commands.add(
                 move |mut idle: CompMut<IdleKickBomb>, mut lit: CompMut<LitKickBomb>| {
                     idle.remove(entity);
-                    lit.insert(entity, LitKickBomb { age: 0.0 });
+                    lit.insert(
+                        entity,
+                        LitKickBomb {
+                            damage_delay: Timer::new(Duration::from_secs_f32(8.0), TimerMode::Once),
```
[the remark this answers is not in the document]
lvllvl: Ok, do you mean 2 timers within the LitKickBomb struct?

So if I understand correctly, we'd end up with

```rust
pub struct LitKickBomb {
     fuse_time: Timer,
     arm_delay: Timer,
}
```

And in initialization of LitKickBomb on line `144` we'd add both fuse_time and arm_delay. 

Ok I'll get this done.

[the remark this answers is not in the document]
zicklag: Yep, that's right!

core/src/elements/kick_bomb.rs
```diff
@@ -177,21 +197,19 @@ fn update_lit_kick_bombs(
...
             explosion_frames,
-            arm_delay,
             fin_anim,
             ..
         } = &element_meta.builtin else {
             unreachable!();
         };
 
-        kick_bomb.age += 1.0 / crate::FPS;
+        kick_bomb.fuse_time.tick(time.delta());
```
zicklag, line 212: We need to make sure we tick both timers here:

```suggestion
        kick_bomb.fuse_time.tick(time.delta());
        kick_bomb.arm_delay.tick(time.delta());        
```

core/src/elements/kick_bomb.rs
```diff
@@ -241,13 +259,13 @@ fn update_lit_kick_bombs(
...
                 body.velocity.y = kick_velocity.y;
-            } else if kick_bomb.age >= *arm_delay {
+            } else if kick_bomb.fuse_time.finished() {
                 should_explode = true;
             }
         }
 
         // If it's time to explode
-        if kick_bomb.age >= *fuse_time || should_explode {
+        if kick_bomb.arm_delay.finished() || kick_bomb.fuse_time.finished() || should_explode {
```
zicklag, line 268: This should only need to check the fuse time here:

```suggestion
        if kick_bomb.fuse_time.finished() || should_explode {
```

zicklag (changes requested):
Got a couple tweaks. And good job keeping comming with this stuff! 🏃

core/src/elements/kick_bomb.rs
```diff
@@ -241,13 +260,13 @@ fn update_lit_kick_bombs(
             } else if !player_standing_left && player_sprite.flip_x {
                 body.velocity.x = -kick_velocity.x;
                 body.velocity.y = kick_velocity.y;
-            } else if kick_bomb.age >= *arm_delay {
+            } else if kick_bomb.fuse_time.finished() {
```
zicklag, line 263: This should be `arm_delay` here:

```suggestion
            } else if kick_bomb.arm_delay.finished() {
```

zicklag (changes requested):

core/src/elements/kick_bomb.rs
lvllvl: Ok just updated this to arm_delay. About to commit right now.

zicklag (changes requested):
And, my bad, there's one last thing, then we'll get this merged.

On [this line](https://github.com/fishfolk/jumpy/blob/26074775f0b709dc2936bcec7760d7d8e11a8554/core/src/metadata/element.rs#LL258C12-L258C12) for fuse_time, and [this line](https://github.com/fishfolk/jumpy/blob/26074775f0b709dc2936bcec7760d7d8e11a8554/core/src/metadata/element.rs#L269) for arm_delay, could you change them to be a `Duration` and add the `serde_humantime` annotation like we have on [this line](https://github.com/fishfolk/jumpy/blob/26074775f0b709dc2936bcec7760d7d8e11a8554/core/src/metadata/element.rs#L284).

And finally update the values in the YAML [here](https://github.com/fishfolk/jumpy/blob/26074775f0b709dc2936bcec7760d7d8e11a8554/assets/elements/item/kick_bomb/kick_bomb.element.yaml#L4) for `fuse_time` ( and the same with `arm_delay` ) to add the `s` suffix, like `8.0s` for 8 seconds or 0.5s for 0.5 seconds.

lvllvl:
> And, my bad, there's one last thing, then we'll get this merged. Sorry!
> 
> On [this line](https://github.com/fishfolk/jumpy/blob/26074775f0b709dc2936bcec7760d7d8e11a8554/core/src/metadata/element.rs#LL258C12-L258C12) for fuse_time, and [this line](https://github.com/fishfolk/jumpy/blob/26074775f0b709dc2936bcec7760d7d8e11a8554/core/src/metadata/element.rs#L269) for arm_delay, could you change them to be a `Duration` and add the `serde_humantime` annotation like we have on [this line](https://github.com/fishfolk/jumpy/blob/26074775f0b709dc2936bcec7760d7d8e11a8554/core/src/metadata/element.rs#L284).
> 
> And finally update the values in the YAML [here](https://github.com/fishfolk/jumpy/blob/26074775f0b709dc2936bcec7760d7d8e11a8554/assets/elements/item/kick_bomb/kick_bomb.element.yaml#L4) for `fuse_time` ( and the same with `arm_delay` ) to add the `s` suffix, like `8.0s` for 8 seconds or 0.5s for 0.5 seconds.

no problem. Let me know if there's anything else I can update. Thanks for the help

core/src/elements/kick_bomb.rs
```diff
@@ -137,7 +144,19 @@ fn update_idle_kick_bombs(
...
+                        entity,
+                        LitKickBomb {
+                            arm_delay: Timer::new(
+                                Duration::from_secs_f32(arm_delay.as_secs_f32()),
+                                TimerMode::Once,
+                            ),
+                            fuse_time: Timer::new(
+                                Duration::from_secs_f32(fuse_time.as_secs_f32()),
+                                TimerMode::Once,
+                            ),
```
zicklag, line 157: This can be simplified slightly:

```suggestion
                            arm_delay: Timer::new(
                                arm_delay,
                                TimerMode::Once,
                            ),
                            fuse_time: Timer::new(
                                fuse_time,
                                TimerMode::Once,
                            ),
```

zicklag (changes requested):
> Thanks for the help

Thank you for the help! 🙂👍

core/src/elements/kick_bomb.rs
lvllvl: Ok this has been simplified / updated. committing now

zicklag (approved):
Great!

PR #758, opened 2023-04-04
refactor: update mine object to use bones time resource instead of manually setting time

lvllvl:
Refactor update for mine object. `thrown_mine` object now uses `arm_delay` of type `Timer`. 

I know we had 2 timers in previous refactors, did we want to do the same here? If so, where would the second timer come from?

### DIFF
--- a/core/src/elements/mine.rs
+++ b/core/src/elements/mine.rs
@@ -13,12 +13,10 @@ pub fn install(session: &mut GameSession) {
 #[ulid = "01GPRSBWQ3X0QJC37BDDQXDNF3"]
 pub struct IdleMine;
 
-// #[derive(Clone, TypeUlid, Debug, Copy)]
 #[derive(Clone, TypeUlid, Debug)]
 #[ulid = "01GPRSBWQ3X0QJC37BDQXDNASF"]
 pub struct ThrownMine {
-    // How long the mine has been thrown.
-    // age: f32,
+    // When this timer reaches the fuse duration, the mine will explode.
     arm_delay: Timer,
 }
 
@@ -130,8 +128,6 @@ fn update_idle_mines(
                 commands.add(
                     move |mut idle: CompMut<IdleMine>, mut thrown: CompMut<ThrownMine>| {
                         idle.remove(entity);
-                        // thrown.insert(entity, ThrownMine { age: 0.0 }); TODO: delete this and next line
-                        //  Timer::new(Duration::from_millis(0), TimerMode::Once),
                         thrown.insert(
                             entity,
                             ThrownMine {
@@ -184,7 +180,6 @@ fn update_thrown_mines(
             explosion_frames,
             explosion_sound,
             explosion_atlas,
-            // arm_delay, TODO: delete this line 
             arm_sound,
             armed_frames,
             armed_fps,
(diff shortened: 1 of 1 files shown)


core/src/elements/mine.rs
```diff
@@ -166,9 +189,9 @@ fn update_thrown_mines(
         };
 
         let frame_time = 1.0 / crate::FPS;
-        thrown_mine.age += 1.0 / crate::FPS;
+        thrown_mine.arm_delay.tick(time.delta());
 
-        if thrown_mine.age >= *arm_delay && thrown_mine.age - *arm_delay < frame_time {
+        if thrown_mine.arm_delay.finished() && thrown_mine.arm_delay.elapsed_secs() < frame_time {
```
zicklag, line 194: `Timer` actually has a helper function that's perfect for this:

```suggestion
        if thrown_mine.arm_delay.just_finished() {
```

core/src/elements/mine.rs
```diff
@@ -12,11 +13,11 @@ pub fn install(session: &mut GameSession) {
 #[ulid = "01GPRSBWQ3X0QJC37BDDQXDNF3"]
 pub struct IdleMine;
 
-#[derive(Clone, TypeUlid, Debug, Copy)]
+#[derive(Clone, TypeUlid, Debug)]
 #[ulid = "01GPRSBWQ3X0QJC37BDQXDNASF"]
 pub struct ThrownMine {
-    // How long the mine has been thrown.
-    age: f32,
+    // When this timer reaches the fuse duration, the mine will explode.
```
zicklag, line 19: The mine actually doesn't explode when this timer finishes, unless it is colliding with a player.

A more accurate description would be something like: “The mine won't explode until this timer finishes.”

zicklag (changes requested):
We don't need two timers in this case because the mine only explodes once a player walks into it.

This looks good other than a couple minor points!

core/src/elements/mine.rs
lvllvl: Ok I made the suggested updates. Thanks for the feedback! Lmk if there's other changes you'd like me to make. 

Btw is working as a software engineer kind of like this? Like working on tickets/issues and resolving them? 

Are the issues generally _always_ harder than `good first issue` tickets?

zicklag: > Btw is working as a software engineer kind of like this? Like working on tickets/issues and resolving them?

Yeah, I think so, though my only software development experience is on GitHub. 😄

> Are the issues generally always harder than good first issue tickets?

Maybe not always, and it depends a lot on how much experience you have with a certain topic.

This kind of stuff gets a lot easier as you continue to do it. Once you've worked in a certain field/language/project long enough you build familiarity with it, and it doesn't take as many tries before you get it right.

I've learned amost everything I know just by trying to make stuff work, and most of the time I didn't know what I was doing when I started. Now I've done that so many times on so many different projects, that a lot of things aren't so hard.

It's cool because the more you practice and work on different things, the easier it gets to do new things you've never done before.

The most imporant thing is to learn how to learn more, then eventually you can do anything.

lvllvl: awesome, thanks for answering my questions.

PR #770, opened 2023-04-22
fix: network UI crashes.

lecoqjacob:
Address a couple network crashes around UI functionality


zicklag (approved):
Awesome, thanks! There's probably quite a few rough edges around error handling so any fixes around that great right now.

PR #771, opened 2023-04-29
refactor: editor to use map_interface

AustinHellerRepo:
The player interactions handled by the editor.rs file have been refactored out into a separate struct so that future enhancements to the editor may use this refactored code instead of copying large sections of it if similar functionality is used (and will be used as part of my next update to allow for map randomization).
An example of how this refactoring will help is when large parts of the map will be updated with one button click. Now simply calling map_interface.set_tile multiple times is possible instead of copy-pasting all of the code previously contained in the SetTile match-case.


zicklag (changes requested):
Awesome, looks great, thanks!

Regarding the name, I think I like `MapManger`, which also has a slight precendent in Jumpy with the `SessionManager` existing already.

Unless you don't like that name, I'm not super picky. :D

Regarding the macro, could you move it to `core/src/utils.rs` ( or maybe `core/src/utils/macros.rs` if you think that seems appropriate ), and then in `core/src/lib.rs` we can add a `#[macro_use]` import for `utils` so that we can use the macro throughout the crate.

Then we can also remove macro definition from `collisions.rs`.

AustinHellerRepo:
I've updated the struct name and centralized the macro. Please let me know if you would like any other changes. Thanks!

zicklag (approved):
Perfect!

PR #773, opened 2023-05-12
fix: fish school removed from editor on element deletion

AustinHellerRepo:
The rendered fish school (and the subsequent fish) will now be removed when the parent element is removed (right-click element and choose "Delete Element").

### DIFF
--- a/core/src/editor.rs
+++ b/core/src/editor.rs
@@ -1,5 +1,4 @@
 use crate::impl_system_param;
-use crate::prelude::fish_school::FishSchool;
 use crate::{map::z_depth_for_map_layer, prelude::*};
 
 pub fn install(session: &mut CoreSession) {
@@ -19,7 +18,7 @@ impl_system_param! {
         tile_layers: CompMut<'a, TileLayer>,
         tiles: CompMut<'a, Tile>,
         tile_collisions: CompMut<'a, TileCollisionKind>,
-        fish_school: CompMut<'a, FishSchool>
+        element_kill_callbacks: Comp<'a, ElementKillCallback>
     }
 }
 
@@ -126,10 +125,10 @@ impl<'a> MapManager<'a> {
         transform.translation.y = position.y;
     }
     fn delete_element(&mut self, entity: Entity) {
-        if let Some(fish_school) = self.fish_school.get_mut(entity) {
-            let kill_callback = fish_school.kill_callback.clone();
+        if let Some(element_kill_callback) = self.element_kill_callbacks.get(entity) {
+            let system = element_kill_callback.system.clone();
             self.commands
-                .add(move |world: &World| (kill_callback.lock().unwrap().run)(world).unwrap());
+                .add(move |world: &World| (system.lock().unwrap().run)(world).unwrap());
         } else {
             // this entity does not contain a kill callback
             self.entities.kill(entity);
--- a/core/src/editor.rs
+++ b/core/src/editor.rs
@@ -1,4 +1,5 @@
 use crate::impl_system_param;
+use crate::prelude::fish_school::FishSchool;
 use crate::{map::z_depth_for_map_layer, prelude::*};
 
 pub fn install(session: &mut CoreSession) {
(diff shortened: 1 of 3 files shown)


core/src/editor.rs
```diff
@@ -124,7 +126,14 @@ impl<'a> MapManager<'a> {
         transform.translation.y = position.y;
     }
     fn delete_element(&mut self, entity: Entity) {
-        self.entities.kill(entity);
+        if let Some(fish_school) = self.fish_school.get_mut(entity) {
```
zicklag, line 129: Awesome!

Can you move the kill callback to it's own component, such as `ElementKillCallback`?

That will let us check whether or not the element being deleted has a kill callback, without making it specific to the fish school.

That way, all we have to do to make this work for, snails, crabs, items, players, etc. will be to add the `ElementKillCallback` component to their elements.

We should probably put `ElementKillCallback` in the `elements.rs` module, and it'd probably be good to give it a `new()` method that takes a generic type implementing `IntoSystem` ( [see here](https://github.com/fishfolk/jumpy/blob/e456f2f88ea44ed3be0c2330ca08b857a23338b3/core/src/item.rs#L197) for how to do the generic arguments ) so that there's less boilerplate when creating an instance of the component.

zicklag (changes requested):

core/src/editor.rs
```diff
@@ -124,7 +125,14 @@ impl<'a> MapManager<'a> {
         transform.translation.y = position.y;
     }
     fn delete_element(&mut self, entity: Entity) {
-        self.entities.kill(entity);
+        if let Some(element_kill_callback) = self.element_kill_callbacks.get(entity) {
+            let system = element_kill_callback.system.clone();
+            self.commands
+                .add(move |world: &World| (system.lock().unwrap().run)(world).unwrap());
```
zicklag, line 131: This was causing a panic because it wasn't actually deleting the element anymore, just the fish.

We could either make the kill callback responsible for actually killing the element entity, or better than that I think, we could just run the kill callback first, and then delete the element entity like this:

```suggestion
            let system = element_kill_callback.system.clone();
            self.commands.add(move |world: &World| {
                // Run the kill callback first
                (system.lock().unwrap().run)(world).unwrap();

                // Then delete the entity
                let entities = world.get_resource::<Entities>().unwrap();
                let mut entities = entities.borrow_mut();
                entities.kill(entity);
            });
```

core/src/elements/fish_school.rs
```diff
@@ -302,3 +312,24 @@ pub fn update_fish_schools(
...
+            .iter()
+            .copied()
+            .for_each(|fish_entity| {
+                to_kill.push(fish_entity);
+            });
+        to_kill.push(entity);
+
+        to_kill.into_iter().for_each(|entity| {
+            entities.kill(entity);
+        });
```
zicklag, line 332: A minor tweak here can avoid an allocation. Often times this wouldn't work because iterating over entities usually needs to borrow `entities`, but in this case it's just a `Vec` so we're good.

```suggestion
        fish_school
            .get_mut(entity)
            .unwrap()
            .fish
            .iter()
            .copied()
            .for_each(|fish_entity| entities.kill(fish_entity));
```

zicklag (changes requested):
Just a couple tweaks. If you've got no other thoughts, you should be able to apply them from the Github UI.

core/src/editor.rs
AustinHellerRepo: This causes an additional search of the entities that would not be necessary if we kill the element as part of its kill callback function. I think it's because your other change removes the kill of the entity.

core/src/elements/fish_school.rs
AustinHellerRepo: We could also kill the parent entity here with another call to `entities.kill(entity)`.

core/src/editor.rs
AustinHellerRepo: I just confirmed the panic is due to the fact that the FishSchool still exists with a vector of Fish entities, but since they have been killed they are no longer returning back the expected Transforms (for moving the fish around). I would think that it would make more sense if the kill callback did all of the work of killing the element (if it exists for the entity). If you really want to kill the parent at the end, maybe we should rename the component to something akin to "dispose callback", "before kill callback", or "prekill callback". Thoughts?

zicklag: > I think it's because your other change removes the kill of the entity.

Oh, yeah, duh. I just missed that, LOL. Good catch. :)

> I would think that it would make more sense if the kill callback did all of the work of killing the element (if it exists for the entity).

If that makes sense to you, I'm fine with that. I could go either way, I just didn't want it to be confusing.

zicklag (approved):
Looks good, thanks!

PR #775, opened 2023-05-13
feat: re-implement blunderbuss.

michanwalsh:
Ok, so far what I've done is add the spread and num_bullets options to the musket metadata, and try to implement some of the logic in musket.rs, specifically by looping through the num_bullets and generating a bullet with a direction that's offset by a random amount of the spread. 

Some questions I guess I'd have are: 
- Where can I find the images/media to use for blunderbass and its effects?
- Should all of the properties I'd need for the new yaml files be contained in the old json? Should I expect to find different variable names for certain properties? Should I just use my best judgement to figure out what corresponds to what?


zicklag:
Here's the [blunderbass](https://github.com/fishfolk/jumpy/blob/main/old_assets/textures/items/Blunderbass(68x24).png) and [blunderbuss](https://github.com/fishfolk/jumpy/blob/main/old_assets/textures/items/Blunderbuss(65x25).png) sprites.

As for the YAML properties, they have changed dramatically, so don't worry too much about lining them up exactly. Really all I think we will need to get the shotgun effect is the `num_bullets` and `spread` options.

Also, I believe our units for speed and such have changed, so you'll need to test locally to see if it's working right.

assets/elements/item/musket/musket.element.yaml
zicklag, line 1: We want to leave this musket file un-changed, since we still want to have the normal Musket.

michanwalsh: Ah yeah I meant to change the Blunderbus copy
Sorry for the late response. I've been trying to run it on my pc, it works but as soon as I add something it's breaking.. so I'm trying to figure out why

zicklag:
I'm going to close this now to tidy things up and since things have changed a bit since this was opened.

For reference, there's now a work-in-progress blunderbass item implemented with the new lua scripting system here: https://github.com/fishfolk/jumpy/tree/main/packs/devpack/items/blunderbass.

And feel free to open a new PR for the blunderbuss if you ever get back to it!

PR #777, opened 2023-05-13
fix: implement kill callback for all rendered elements

AustinHellerRepo:
All elements that act as a spawner (except for the fish school since the spawner acts as the parent element holding the inner fish elements) now have a Spawner component that will be discovered during the "delete element" process.
I also updated the ElementKillCallback ulid string since I just created it manually in the previous ticket and only recently learned of a way to generate a true random ulid string.
We should consider refactoring the fish school spawner to properly behave like a spawner containing entities.

### DIFF
--- a/core/src/editor.rs
+++ b/core/src/editor.rs
@@ -18,7 +18,8 @@ impl_system_param! {
         tile_layers: CompMut<'a, TileLayer>,
         tiles: CompMut<'a, Tile>,
         tile_collisions: CompMut<'a, TileCollisionKind>,
-        element_kill_callbacks: Comp<'a, ElementKillCallback>
+        element_kill_callbacks: Comp<'a, ElementKillCallback>,
+        spawners: Comp<'a, Spawner>
     }
 }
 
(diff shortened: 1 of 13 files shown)


zicklag:
Yeah, I'm thinking maybe we don't need the `ElementKillCallback`, maybe, if we have the `Spawner` abstraction.

The fish school could easily replace it's vector of fish with the entities from the spawner with no negative side-effects, and then we'd no longer have anything using the `ElementKillCallback`, so maybe it'd be better just to get rid of it.

What do you think?

AustinHellerRepo:
> Yeah, I'm thinking maybe we don't need the `ElementKillCallback`, maybe, if we have the `Spawner` abstraction.
> 
> The fish school could easily replace it's vector of fish with the entities from the spawner with no negative side-effects, and then we'd no longer have anything using the `ElementKillCallback`, so maybe it'd be better just to get rid of it.
> 
> What do you think?

I completely agree. It's a nice custom feature to have, but if nothing is using it then we're just going to end up maintaining this feature through future maintenance and functionality, maybe even having to put in workarounds, just for this unused feature to exist. I'll refactor this branch to include the fish school's fish and then remove the `ElementKillCallback`. Thoughts?

zicklag:
Sounds like a plan!

zicklag (approved):
Looks good to me, is this ready?

AustinHellerRepo:
Yeah, this is good for now. The rendered players still don't get removed yet, but that's something that might be a little bigger and different from this PR. I'll tackle that one next.

PR #779, opened 2023-05-19
Fix player spawner kill

AustinHellerRepo:
The player spawners can now be removed and the last player spawner to be removed will also remove the rendered players. This is done because the players are able to spawn from a single player spawner (if that is the only spawner), so removing the rendered players after removing all but the last spawner would simply cause them to respawn from the existing player spawner(s).

The ElementKillCallback needed to be brought back into the system in order to facilitate the complex killing of the rendered player entities.

### DIFF
--- a/core/src/editor.rs
+++ b/core/src/editor.rs
@@ -133,17 +133,15 @@ impl<'a> MapManager<'a> {
         } else {
             if let Some(spawner) = self.spawners.get(entity) {
                 // search for other spawners in the same group
-                let is_last_spawner_from_group = self
+                let grouped_spawners_count = self
                     .spawners
                     .iter()
                     .filter(|other_spawner| {
                         spawner.group_identifier == other_spawner.group_identifier
                     })
-                    .peekable()
-                    .peek()
-                    .is_some();
+                    .count();
 
-                if is_last_spawner_from_group {
+                if grouped_spawners_count == 1 {
                     spawner.spawned_elements.iter().copied().for_each(|entity| {
                         // TODO recursively search for nested spawners
 
--- a/core/src/editor.rs
+++ b/core/src/editor.rs
@@ -130,22 +130,30 @@ impl<'a> MapManager<'a> {
             let system = element_kill_callback.system.clone();
             self.commands
                 .add(move |world: &World| (system.lock().unwrap().run)(world).unwrap());
-        }
-        else {
+        } else {
             if let Some(spawner) = self.spawners.get(entity) {
                 // search for other spawners in the same group
-                let is_last_spawner_from_group = self.spawners.iter().filter(|other_spawner| spawner.group_identifier == other_spawner.group_identifier).count() == 1;
+                let is_last_spawner_from_group = self
+                    .spawners
+                    .iter()
+                    .filter(|other_spawner| {
+                        spawner.group_identifier == other_spawner.group_identifier
+                    })
+                    .peekable()
+                    .peek()
+                    .is_some();
 
                 if is_last_spawner_from_group {
                     spawner.spawned_elements.iter().copied().for_each(|entity| {
                         // TODO recursively search for nested spawners
 
-                        if let Some(element_kill_callback) = self.element_kill_callbacks.get(entity) {
+                        if let Some(element_kill_callback) = self.element_kill_callbacks.get(entity)
+                        {
                             let system = element_kill_callback.system.clone();
-                            self.commands
-                                .add(move |world: &World| (system.lock().unwrap().run)(world).unwrap());
-                        }
-                        else {
+                            self.commands.add(move |world: &World| {
+                                (system.lock().unwrap().run)(world).unwrap()
+                            });
+                        } else {
                             self.entities.kill(entity);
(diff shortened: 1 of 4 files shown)


zicklag (commented):
This looks great, tested it and works.

BTW having everything delete properly when you are in the editor makes it feel much more polished. Thanks for putting that in!

My only thought is that it might be cleaner to use a resource to keep track of all the player spawners and which players are spawned, instead of kind of synchornizing the data between spawners.

That might allow us to get rid of the group_identifier in the spawner too.

Just to outline what I'm thinking:

- Store the list of spawned players, and the list of spawners in a resouce instead.
- Whenever a spawner is deleted, it can use the `ElementKillCallback` to remove itself from the list of spawners.
    - If it is the last spawner, then it can delete all the players in the spawned player list.
- Whenever a player is spawned or killed, it is added or removed from the list of spawned players in the resource.

What do you think?

AustinHellerRepo:
> This looks great, tested it and works.
> 
> BTW having everything delete properly when you are in the editor makes it feel much more polished. Thanks for putting that in!
> 
> My only thought is that it might be cleaner to use a resource to keep track of all the player spawners and which players are spawned, instead of kind of synchornizing the data between spawners.
> 
> That might allow us to get rid of the group_identifier in the spawner too.
> 
> Just to outline what I'm thinking:
> 
>     * Store the list of spawned players, and the list of spawners in a resouce instead.
> 
>     * Whenever a spawner is deleted, it can use the `ElementKillCallback` to remove itself from the list of spawners.
>       
>       * If it is the last spawner, then it can delete all the players in the spawned player list.
> 
>     * Whenever a player is spawned or killed, it is added or removed from the list of spawned players in the resource.
> 
> 
> What do you think?

I'm not sure that we should deviate from the Spawner logic such that there is a special case for the PlayerSpawner. We should probably aim to fix the implementation to be less messy for grouped spawners. Maybe it would be better if the logic wasn't so elaborate in the `player_spawner.rs` file and instead we had two static functions on the `Spawner` struct that would insert the "ungrouped" or "grouped" spawner based on the respective function.
For example:
```
impl Spawner {
    pub fn insert_grouped_spawner<T>(entity, spawned_elements, &mut entities, &mut spawners, elements: CompMut<T>) {
        // search entities similar to the logic below
                    // try to find one other spawner and store the existing player entities
            if let Some((_, (_, first_spawner))) =
                entities.iter_with((&elements, &spawners)).next()
            {
                // all of the player spawners share the same group identifier
                let spawner = Spawner::new_grouped(
                    first_spawner.spawned_elements.clone(),
                    first_spawner.group_identifier.clone(),
                );
                spawners.insert(entity, spawner);
            } else {
                let spawner = Spawner::new(vec![]);
                spawners.insert(entity, spawner);
            }
    }
    pub fn insert_ungrouped_spawner(entity, spawned_elements, &mut entities, &mut spawners) {
        let spawner = Spawner::new(spawned_elements);
        spawners.insert(entity, spawner);
    }
}
```
Thoughts?

zicklag:
I like the idea of moving the complication into the spawner as much as possible instead.

I think the piece that bugs my "OCD" the most is the fact that we're kind of synchronizing the player entity list with all the player spawners, when that information would ideally be stored in one place. That's why I was thinking of using a resource to store the lists in one place.

Maybe each `Spawner` _component_ just contains a spawner group ID, and then we have a `SpawnerEntities` _resource_, that stores like a `HashMap<String, Vec<Entities>>`, mapping group ids to the entity list. That way, even if you have a bunch of spawners with the same group ID, the entities are only stored in one place.

If you think that's over-thought, though, I'm also fine just merging this as-is.

AustinHellerRepo:
That's a great idea. I agree with you that having (what is basically) duplicate data across many different components is not ideal. I like your idea of storing it off in a resource as a solution to this problem. Thank you for suggesting it. I'll make those changes and will push out a new PR soon.

AustinHellerRepo:
As of commit c64bf2a, the spawner logic is managed by the SpawnerManager.
The following is now possible:
* the spawner can be killed via a one-liner (but advised to check the type in some cases)
* the entities of any spawner are not contained in the SpawnerEntities resource (instead of copied between grouped spawners)
* the creation of a spawner entity (and any potential default spawned entities) can now be stored via a one-liner, either grouped or ungrouped
* new spawned entities can now be pushed into or removed from the SpawnerEntities resource (ultimately) via a one-liner

zicklag (approved):
_Repeating my reply to your question on naming in Discord:_

> Do you have any opinions about the function names of the SpawnerManager? I'm not sure that I like them. I'm thinking instead of insert_spawned_entity_into_grouped_spawner it could be create_grouped_spawner (which accepts an initial vector of spawned entities).

Yeah, I think that sounds a little better.

Also, I was thinking maybe we don't need to say ungrouped for all the ungrouped functions, and just treat that as the "default", but make sure all the grouped ones have grouped in the name.

Otherwise, I think I like the latest version is looking good :thumbsup:

zicklag (approved):
Awesome, thank you!

PR #784, opened 2023-05-25
build: remove transitive bevy_pbr dependency.

### LINKED ISSUE #614: Chore: Figure Out Why We Depend on `bevy_pbr`.
Somehow Jumpy is depending on the `bevy_pbr` crate, which we shouldn't need. We should find out how to disable that.

tzemanovic:
closes #614 

Removes transitive bevy_pbr dependency as found with help of `cargo tree -i bevy_pbr `.


zicklag (approved):
Perfect, thank you!

PR #789, opened 2023-05-31
feat: add randomization feature to map editor.

AustinHellerRepo:
A "randomize" button now exists in the map editor which will systematically find groups of tiles and elements, shifts them around randomly, and stops once it finds that none of the groups overlap.


zicklag (commented):
Just tested and it works! I noticed more dependencies than seemed right, but I think I might have found the culprit: https://github.com/AustinHellerRepo/Shiftnanigans/issues/2.

I also resolved conflicts.

Now I'm going to check out the code.

core/src/map_constructor.rs
```diff
@@ -0,0 +1,7 @@
+use crate::editor::MapManager;
+
+pub mod shiftnanigans;
+
+pub trait MapConstructor {
+    fn construct_map(&self, map_manager: &mut MapManager);
```
zicklag, line 6: I like the idea of making a trait out of this. I could definitely see having different map generation techniques eventually. For example, something like [mario-gpt](https://github.com/shyamsn97/mario-gpt) has already been brought up.

zicklag (approved):
OK, this is looking good!

I notice that on certain maps it maxes out the CPU and freezes for an extended period of time, at which point the window manager says that it's not responding, but I think this is fine for MVP. I'll open an issue so we don't forget about it.

I opened https://github.com/AustinHellerRepo/Shiftnanigans/pull/3 to shave a few dependencies off, but I think the change is non-breaking, so I can merge this how it is, and we can pick up the update with a just `cargo update` once that's ready.

Awesome work!

core/src/map_constructor.rs
AustinHellerRepo: That was exactly what I was thinking too, which is why I was wanting to ensure that the `MapManager` could handle everything necessary for working with tiles/elements in the editor. I'm so glad that it's finally done and working well. I'll take a swing at that opened issue in Shiftnanigans next and then I'll work through the occasional freezes after that. Thank you for all of your help. I really appreciate it.

zicklag: I'm so glad I can help! Thank you for improving our game! :heart:

Issue #792, opened 2023-06-02
Prune Dependencies Where Possible


zicklag:
We're currently running at ~581 dependencies for a build of Jumpy on Linux, which is quite a lot.

Lots of it is unavoidable, but we also have several duplicate dependencies detected by `cargo deny check ban`, which we might be able to eliminate.

We should try to prune our dependency tree as much as we can without causing adverse side-effects, just to keep things tidy as possible, and to reduce clean compile times, which can sometimes be off-putting to new contributors, and at the very least, add lead-in time when first jumping into jumpy development :wink:.

orhun:
Hey, have you tried running [`cargo-udeps`](https://github.com/est31/cargo-udeps)?

zicklag:
Good idea, I hadn't run it recently. I just tried it now and it says all dependencies appear to be used.

zicklag:
Ah, `cargo machete` found some "unused" deps, though. They were technically used somewhere in our dependency tree, but not directly used by the crate itself.

So it didn't help remove dependencies per-se, but it did find entries that we could harmlessly remove from our `Cargo.toml`.

orhun:
wow, TIL `cargo-machete`

PR #793, opened 2023-06-02
build: cargo update and remove unused deps.

zicklag:
Also updates to use two digit version numbers for most crates.


zicklag:
Works on #792 but doesn't completely solve the issue.

Issue #714, opened 2023-03-14
Fix Swords Killing Yourself Too Easily


zicklag:

zicklag:
Looks like it might not be happening anymore, anyone can comment if they notice this issue again and I'll re-open.

PR #809, opened 2023-06-16
fix: input controls being applied to ai players.

### LINKED ISSUE #722: Second Player Dies When Pressing Right Shift in Some Circumstances
@rafaelcastrocouto#0546 on discord reported that this key combination would kill the second player for some reason with the default key bindings:

```
Key press: Some(RShift) (54)
Key release: Some(RShift) (54)
Key release: None (0)
```

tzemanovic:
fixes #722 

I'm still getting familiar with the code, but it looks like what's happening is that when you press the Grab/Drop key on Keyboard 2, the input sometimes makes the AI player drop the sword and when it has enough velocity it gets insta-killed by collision detection (this I think is related to #714). I was able to reproduce this issue in a local game only.

After some digging, I found that this seems to be caused the session's input being propagated to all the player inputs including the AIs in a local session (when `network_player_idx.is_some()`). Removing the condition skips the loop for AI players locally and still works in a network game, where each local player has `player_idx = 0`.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -266,7 +266,7 @@ fn collect_local_input(
     }
 
     for (player_idx, action_state) in &player_input_collectors {
-        if network_player_idx.is_some() && player_idx.0 != 0 {
+        if player_idx.0 != 0 {
             continue;
         }
 


zicklag:
Awesome, thank you! I'll try to check this out when I get the chance.

@lecoqjacob I requested your review, in case you want to and you can get to checking this out before me.

lecoqjacob:
In the scenario where we have AI players in an online session, does this break that rule?

zicklag:
We actually don't let you have AI in an online match at the moment. Still we'll probably want that so we might want to code defensively and assume that it could happen.

tzemanovic:
Yeah, I think this will still work with AI players in an online session because their input is being controlled elsewhere (`fn player_ai_system`)

zicklag:
Sounds good, mering! Thank you!

PR #820, opened 2023-06-20
feat: add coral spikes

Zac8668:
Adds a simple element that kills the player in contact, called Spike.


core/src/elements/spike.rs
```diff
@@ -0,0 +1,88 @@
...
+        if let BuiltinElementKind::Spike {
+            atlas,
+            body_size,
+            start_frame,
+            end_frame,
+            fps,
+            ..
+        } = &element_meta.builtin
+        {
+            new_spikes.push(entity);
```
zicklag, line 44: This `new_spikes` vector looks unused.

zicklag (changes requested):
One tweak, but this looks good. This should be really useful for map design.

zicklag (approved):

PR #823, opened 2023-06-22
fix: improve menu flow with escape action.

lecoqjacob:


zicklag (approved):
This was bugging me, thanks for fixing!

PR #824, opened 2023-06-22
fix: adjust emote region to have an owner with a buffer

lecoqjacob:


lecoqjacob:
Peek 2023-06-22 10-40

assets/player/skins/fishy/fishy.player.yaml
```diff
@@ -234,9 +234,6 @@ layers:
       emote_alarm:
         fps: *fps
         frames:
-          - 0
-          - 0
-          - 0
```
zicklag, line 239: we should remove these for the other players too.

zicklag (changes requested):
Looks good, just have to remove the padding frames from the other player skins.

assets/player/skins/fishy/fishy.player.yaml
lecoqjacob: Removed!

lecoqjacob:
@zicklag removed the remaining buffers. Added the emote region to the bullet. We should probably extend the x value for the bullet emote region. It moves pretty fast and we want the players to emote a bit before it hits.

I was also noticing that we can only create rectangle regions for emote regions. Might look into how to create an anchor point so that the region can instead start at the very end of the bullet and not just surrounding it.

image

zicklag:
Sounds like a good idea for the bullet region's offset. Do you want to try that before merging this, or should I just get this in and you can follow-up?

lecoqjacob:
> Sounds like a good idea for the bullet region's offset. Do you want to try that before merging this, or should I just get this in and you can follow-up?

Merge this and ill create a ticket for follow up

zicklag (approved):

PR #829, opened 2023-07-07
feat: add in-game console displaying logs.

MaxCWhitehead:
### **Overview**
Adds console using bevy_console crate, likely should implement our own console down the line. Can be open with grvave key: `

bevy_log::LogPlugin does not allow us to specify a tracing layer to write logs to console, so I duplicated this plugin and modified slightly. There are a number of feature flags for bevy_log that are commented our here - I do not believe these should impact us. We should switch back to bevy's plugin if this feature is added.

I tested on wasm, logging still works (in browser console), however console is disabled in wasm due to limitations of bevy_console. Long term implementing our own console is a good idea for wasm support, hopefully color support, and additional control.

We can register console commands as seen in bevy_console docs, may be useful for cheats, debug commands, etc.

### **Concerns:**
This change ended up being maybe a bit more complex than I hoped, I tried to keep it simple, but only so much can be done there. I am a bit worried about performance issues that may come from this, I have ideas if we find issues. I couldn't get profiling to work, wondering if there is any noticeable impact during heavy logging. Perhaps this can be disabled in release build, but it could impact performance of dev builds and impact profiling results.

### **Regarding performance**
- Log buffer is a simple Vec in a mutex: Right now if the console system is consuming logs for rendering to console, this will block tracing on other threads. We may consider using a double buffer or something to minimize blocking on tracing.
- We are currently cloning an Arc on every trace due to having to use MakeWriter for tracing layer (see comments in change). This seems maybe bad but also hard to say until profiled, conceptually I know this can be fixed but I ran into issues with self-referential structs in Rust.

I opted not to make things complicated by addressing those issues and prematurely optimizing lol. No worries if this change seems like too much or inappropriate, I just thought it'd be nice to have.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -194,55 +194,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "anstream"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0ca84f3628370c59db74ee214b3263d58f9aadd9b4fe7e711fd87dc452b7f163"
-dependencies = [
- "anstyle",
- "anstyle-parse",
- "anstyle-query",
- "anstyle-wincon",
- "colorchoice",
- "is-terminal",
- "utf8parse",
-]
-
-[[package]]
-name = "anstyle"
-version = "1.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "41ed9a86bf92ae6580e0a31281f65a1b1d867c0cc68d5346e2ae128dddfa6a7d"
-
-[[package]]
-name = "anstyle-parse"
-version = "0.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e765fd216e48e067936442276d1d57399e37bce53c264d6fefbe298080cb57ee"
-dependencies = [
- "utf8parse",
-]
-
-[[package]]
-name = "anstyle-query"
-version = "1.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


zicklag:
Big thanks for this @MaxCWhitehead! I pushed a couple commits to your branch:

- The first one just uses `bevy::utils` module instead of adding the `bevy_utils` crate to the `Cargo.toml`.
- The second one simplifies the console log message sending and hopefully avoids any performance problems.

To avoid blocking/locking on the message sending, I used a channel to send messages. In my initial iteration I was thinking of sending `Vec<u8>`s over the channel, but that would allocate for every message which isn't cool.

I did some looking around the Rust crate ecosystem for something that would help us out and finally found `byte-pool`, which allows us to allocate chunks of bytes out of a pool, so that we we don't have to allocate a new `Vec` every time we want to send a message. I think I turned out good, but I'd like to hear your thoughts before I merge.

Thanks again!

MaxCWhitehead:
Thanks a bunch for fixing this up! I feel a lot better about this. byte_pool + Channel seems like a much better approach. Looks great to me, committed a fix for docs.

MaxCWhitehead:
Got console working on wasm. There is a minor bug - console window slowly expands downwards. Guessing a bug in bevy-console, not worried about it is functional.

MaxCWhitehead:
Ah apparently it is a tracked issue - once we merge this will probably link it over there as a repro.

zicklag (approved):
Awesome, this all looks good!

There's a CI failure on the advisory checks, which I'll look at tomorrow probably ( or you can check out if you want, we run [`cargo deny`](https://github.com/EmbarkStudios/cargo-deny) to check our dependencies ).

PR #832, opened 2023-07-08
fix:  Fix puffin profiler not showing anything

### LINKED ISSUE #831: Puffin Profiler Turned Off
### Description

When starting the puffin profiler in the game it says:

```
The puffin profiler is OFF!
No profiling data.
```

### To Reproduce

Start the game and press `f7` to show the profiling window.

### Expected Behavior

The profiling window should show the profiling data being recorded.

### Additional Context

This used to work, so I'm not sure what changed. Maybe we update the puffin version or something.

You're supposed to use `puffin::set_scopes_on(true)` in order to start the profiler, and we do that here, where we turn the scopes on, but only if the window is showing.

https://github.com/fishfolk/jumpy/blob/241d080be6e6a80f4d283d4a31633826ac1dd242/src/ui/debug_tools.rs#L248

I tested just changing that to `true` instead of `show.profiler`, but it's still not working for some reason.

### Log Messages

_No response_

MaxCWhitehead:
Fix #831, we are using incompatible puffin / puffin-egui versions.

We updated puffin to 0.16, but did not upgrade puffin_egui to 0.22 (these versions were bumped together). We likely did not upgrade puffin_egui because this causes a compilation error, as bevy_egui is still using dep egui 0.21 (while puffin_egui is at 0.22), types mismatch.

Downgrading puffin to 0.15 fixes it. Can upgrade puffin crates once bevy_egui has a release with egui 0.22.

Additionally, fixes a panic opening profiler in wasm due to missing web feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2894,7 +2894,7 @@ dependencies = [
  "peg",
  "ping-rs",
  "postcard",
- "puffin 0.15.0",
+ "puffin",
  "puffin_egui",
  "quinn",
  "quinn_runtime_bevy",
@@ -2928,7 +2928,7 @@ dependencies = [
  "nalgebra",
  "ordered-float",
  "petgraph",
- "puffin 0.16.0",
+ "puffin",
  "rapier2d",
  "serde",
  "shiftnanigans",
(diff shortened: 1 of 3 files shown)


MaxCWhitehead:
After another look, not 100% this is fixed, don't seem to be many scopes at all in profiler, gonna poke around.

MaxCWhitehead:
Seeing scopes now - realized most of them were in jumpy_core and did not update the version there.

zicklag:
Awesome, thanks for looking into this!

PR #844, opened 2023-07-19
fix(time_step): Fix time_factor causing velocity to be dependent on crate::FPS.

MaxCWhitehead:
While doing some testing, I changed crate::FPS to be 30 instead of 60, this caused everything to move half as fast.

I do not believe the time needs to be relative to FPS here. The time used is fixed value computed based on timestep. Time resource used here is advanced based on crate::FPS and NETWORK_FRAME_RATE_FACTOR (in networked game, or just crate::FPS in offline). This means it is already adjusted based on frequency of updates. Due to this, objects already move a consistent speed when using different frame rates.

Changing this to scale by 60 makes velocities consistent regardless of crate::FPS value, we are keeping scale by constant 60 to avoid changing velocities everywhere else.

### DIFF
--- a/core/src/physics.rs
+++ b/core/src/physics.rs
@@ -135,13 +135,10 @@ fn update_kinematic_bodies(
 ) {
     puffin::profile_function!();
 
-    // TODO: By default we assume a frame rate of of crate::FPS ( 60 frames-per-second ), but we may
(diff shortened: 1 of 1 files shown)


zicklag (approved):
Good find! I just realized how stupid I was writing `dt / ( 1 / fps )` when it should have been `dt * fps` LOL.

PR #845, opened 2023-07-26
feat(network-debug): Network debug tool shows ggrs network stats per player

MaxCWhitehead:
GGRS has api exposing network stats per remote player, this change collects this data and displays in network debug tool.

This shows current ping, send queue length, estimated frames ahead/behind this player is from local player, and kb/s being sent.

image


MaxCWhitehead:
This data is exposed only for current frame, but may consider buffering this and allowing hover over past frames to display historical data, not sure if this would be useful. I do think I will at least buffer ping at some point and add a graph of ping over time.

Issue #847, opened 2023-07-27
Performance: Skip work in Update Kinematic Bodies if object has not moved since last update.


MaxCWhitehead:
### Description

I setup a map in which I am only player (no AI), and removed critters such that nothing is actively moving.

It seems items and other kinematics are going down paths for shoving out of objects, and checking if should fall through ground running collision detection. Some bodies do check falling path and take 100ms, others seem to do shove -> check falling for 50ms each, cumulative cost seems to be 100ms consistently per object.

The 50ms cost of shove out of objects is just for collision detection, is higher if actually shoving out.

I am interested in exploring if we can "sleep" items that are not moving or interacted with. Basically if gravity is the only force acted on them, we know they should end up in the same spot as last frame, and can skip collision tests with static objects such as tilemap. 

Probably should proceed with care, as these types of sleeping optimizations are easy to do wrong / get bugs, but I believe we should be able to shave something close to 1.5ms off of each game update. Possibly more in a map with more items.

Example of update loop with no critters, one player not moving, and items laying around map:
    
<img width="1249" alt="image" src="https://github.com/fishfolk/jumpy/assets/35712032/d6f6e643-060b-4ad8-ae45-b9d161ed5848">

I expect this should have an impact on #846

zicklag:
I have been worried about the inefficiency of our "shove out of things" loop. I didn't want to pre-maturely optimize, but it felt inefficient.

I wonder if we could take better advantage of things built-into rapier, like maybe the [broad phase](https://docs.rs/rapier2d/latest/rapier2d/geometry/struct.BroadPhase.html) for avoiding searching for collisions with objects not next to the object of interest.

Sleeping does sound like a good idea, too, as long as we can come up with a sensible algorithm for how to wake them up.

MaxCWhitehead:
[...]

PR #848, opened 2023-07-28
feat: Skip update_kinematic_bodies shove loop and check falling branch if object has not moved since last update.

MaxCWhitehead:
So the stat shown in #847 wasn't too representative of the costs as I had removed the critters and such.

This change checks if kinematic has moved since last update, and only tries shove out if it has moved. (If hasn't moved, either isn't colliding and doesn't need shove out, or is stuck and another shove out will yield same results and remain stuck).

This removed about 1ms from update_kinematic_bodies.
image

Would like to leave #847 open, I am aiming to also skip the fallthrough collision check if hasn't moved, however there is additional complexity here related to fall_through bool, and I don't want to break stuff so that will come later. If we can skip that - should be able to get this function to almost only spend time on moving objects and save even more.

### DIFF
--- a/core/src/physics.rs
+++ b/core/src/physics.rs
@@ -80,6 +80,12 @@ pub struct KinematicBody {
     /// This is important to make sure that it falls through JumpThrough platforms if it happens to
     /// spawn inside of one.
     pub is_spawning: bool,
+
+    /// Position cached from last kinematic body update, used to determine if object is "sleeping"
+    /// (is not moving) to avoid collision detection / resolution against static objects.
+    pub last_update_position: Vec2,
+    /// See comment for `last_update_position`, this tracks previous rotation to detect if object has moved.
+    pub last_update_rotation: f32,
 }
 
 impl KinematicBody {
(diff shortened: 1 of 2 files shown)


MaxCWhitehead:
It seems like the check fall through is necessary on things that haven't moved, however, this value is only ever set on player (or things that are controlled), other than in this collision check. This means we should be able to skip check fall through if it has not moved and is not controlled. Will explore and open another PR soon.

MaxCWhitehead:
With 2nd commit, now skipping fall through check too... Things are getting faster.

image

MaxCWhitehead:
I realized these pics were with standard build not optimized, regardless the relative times here seem to be a great improvement.

core/src/physics.rs
```diff
@@ -238,17 +253,24 @@ fn update_kinematic_bodies(
...
-                        .map(|x| *x == TileCollisionKind::JumpThrough)
-                        .unwrap_or(false)
-                }) == TileCollisionKind::JumpThrough
-            {
-                body.fall_through = true;
+            // If not moving, this collision test should give the same result, and will not change the value of fall_through.
+            // for players, fall_through may be modified based on inputs, and we may actually want this to be set again here,
+            // so only skip if not moving for bodies that are not players.
+            let is_player = player_states.get(entity).is_some();
+            if has_moved || is_player {
```
zicklag, line 260: This `is_player` check is something I could see being a footgun for other player-controlled items that may be created later.

I'm open to other ideas, but what if we added a flag to the kinematic body like `is_controlled`, with a note that indicates how it effects sleeping.

Then we just spawn players with their kinematics bodies `is_controlled: true`, and any other player-controlled items can use that too.

MaxCWhitehead: This all sounds good to me.

MaxCWhitehead:
Added a commit making this change, is_controlled is false by default, and is set on player/AI kinematic bodies.

Can ignore the force push, accidentally set is_controlled on the sword attachment for AI.

zicklag (approved):

PR #863, opened 2023-10-19
feat: create `LuaEngine` resource that can be used to run lua systems.

zicklag:
This doesn't include any lua bindings for the World yet, though, so scripts can't do anything but log messages.


zicklag:
Superceeded by scripting support in bones.

Issue #882, opened 2023-12-17
AI clipping into collision boxes and losing control of the player


Biped-Potato:
### Description

Context: The player controlled character is "Fishy" the orange one, every other entity is an AI
1. AI Clipping issue
The AI will occasionally clip into the ground and start repeatedly playing the jump sound resulting in an obnoxious jittering noise while creating a motion as seen in the video
2. Player Control Loss
The ability to control the player is often lost for a period of time where it will just move to the right(i've only experienced to the right). You can stop this movement but not move in the other direction by pressing the left key.

https://github.com/fishfolk/jumpy/assets/79812179/7c41172a-8a1d-4d84-be5d-ce16ee7f0d70

### To Reproduce

After moving around for a while these errors both occur for me

zicklag:
Thanks for the report!

Just for reference, I believe that the issue with the AI getting stuck is a physics issue related to https://github.com/fishfolk/jumpy/issues/853 and the player issue will require more investigation.

No need for action on your part, I'll probably tackle this after getting networking back into MVP state.

zicklag:
I figured out the input issue! I'm pretty sure, anyway, let me know if you have issues after the next release. I know I fixed _something_, and I'm pretty sure it helps a lot with the controls. Random inputs were being dropped or delayed 'till the next frame because of a missing ordering constraint in the Bevy system.

Fix is incoming, and I'll publish a bugfix release that includes both of the latest fixes.

Issue #853, opened 2023-10-14
Objects Get Stuck in Walls, Especially Due to Rotation


zicklag:
### Description

Sometimes objects will get stuck in walls, and just hang there. I don't think this used to happen like this, so I believe it was probably introduced in https://github.com/fishfolk/jumpy/pull/848, which avoided pushing things out of walls as often for performance's sake.

It is also possible to, very rarely, have the player get stuck in the floor, making it impossible to move.

### To Reproduce

An easy way to reproduce is to walk up to a wall and drop a musket while facing the wall.

### Expected Behavior

Items should get pushed out of walls by the physics system.

MaxCWhitehead:
@zicklag I think this should be resolved, but can verify it looks as you expect / close this out.

zicklag:
Yep, looks like it's working now!

PR #883, opened 2023-12-19
fix(character_movement): Fix sleeping kinematic / has_moved optimization causing AI to sink into ground

MaxCWhitehead:
I seem to have set the update of saved positions used to determine if kinematic has moved in the wrong spot, ended up in a situation in which AI thinks it had not moved, but was inching down by -0.1 each frame, sinking into ground.

I was able to repro fairly consistently and feeling pretty confident this should fix the AI sinking into ground described in #882. (Though I do not believe this is related to the losing control of player bit).

EDIT:  #853 (objects stuck in walls) also appears to be resolved by this fix.

### DIFF
--- a/src/core/physics.rs
+++ b/src/core/physics.rs
@@ -173,6 +173,11 @@ fn update_kinematic_bodies(
                 || body.last_update_rotation != rotation
                 || body.is_spawning // Don't consider new objects
         };
+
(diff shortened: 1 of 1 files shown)


zicklag:
Awesome, thanks!

PR #889, opened 2023-12-22
chore: update Cargo.lock to include correct crate version.

zicklag:
This was accidentally missed from the latest commit.


orhun:
we can potentially build with `--locked` flag to avoid situations like this

zicklag:
We actually do build with the `--locked` flag already, so I probably didn't need to do this before tagging the release, but I figured I'd do it just in case and to make things tidy.

PR #891, opened 2023-12-23
chore(build): use shadow-rs for build info

### LINKED ISSUE #890: Allow compilation without git binary
### Description

#887 made the compilation hardly depend on the `git` binary. This blocks me from updating the Arch Linux package (since I'm not using the git repo for building, but the source archives):

```sh
error: git describe exited with status 128: fatal: not a git repository (or any of the parent directories): .git
   --> src/ui/main_menu.rs:100:57
    |
100 |                         egui::TextEdit::singleline(&mut git_version::git_version!())
    |                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this error originates in the macro `git_version::git_version` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `jumpy` (bin "jumpy") due to previous error
```

This can be verified from the [`git-version` documentation:](https://docs.rs/git-version/0.3.9/git_version/)

> These macros do not depend on libgit, but simply uses the git binary directly. So you must have git installed somewhere in your PATH.

### To Reproduce

Build from source without Git.

### Expected Behavior

Successful build.

### Additional Context

I recommend using [`shadow-rs`](https://crates.io/crates/shadow-rs) to conditionally embed the build info instead (in `build.rs`). e.g. use the git version if the git is installed, use the actual version otherwise.

### Log Messages

Shared above.

orhun:
closes #890

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1388,6 +1388,32 @@ dependencies = [
 ]
 
 [[package]]
+name = "const_fn"
+version = "0.4.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fbdcdcb6d86f71c5e97409ad45898af11cbc995b4ee8112d59095a28d376c935"
+
+[[package]]
+name = "const_format"
+version = "0.2.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e3a214c7af3d04997541b18d432afaff4c455e79e2029079647e72fc2bd27673"
+dependencies = [
+ "const_format_proc_macros",
+]
+
+[[package]]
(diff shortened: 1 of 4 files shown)


zicklag (approved):
The `VERSION` constant was apparently displaying multiple versions in like a YAML format, and also needed to be tweaked to return a `&str` so egui knows not to make it look like an editable textbox. I also needed to remove the default features in the non-build dependency so that it would build on WASM, but now everything looks good!

I also tested locally to make sure you could build with just a source snapshot and no git repository.

zicklag:
Published v0.9.3 with this update.

orhun:
That's great, thank you!

PR #896, opened 2024-01-07
fix: Kick Bomb can't be used

nelson137:
While working on the jellyfish I noticed that the kick bomb wasn't working.

When spawned they didn't get an `ElementHandle` component (or `MapElementHydrated`). Without an element handle the `update_idle_kick_bombs` system always skipped over them, preventing them from being used.


zicklag (approved):
Awesome, thanks! I must have messed that up earlier. 😄

PR #897, opened 2024-01-10
feat: Add a find helper method to `PlayerInventories`

nelson137:
I noticed that in many places in the codebase, it's necessary to find the inventory that contains some item (usually to get the player entity), if any exists, from the `PlayerInventories` system param. This simply adds a helper method `find_item` that does the `.iter().find_map(...)` that is duplicated in many places.

There is also some minor cleanup regarding `ComponentStore::remove`.


zicklag (approved):
Looks good!

PR #898, opened 2024-01-10
feat: Jellyfish v2

### LINKED ISSUE #577: Re-implement Jellyfish

nelson137:
Take 2 🎬 this is the second iteration of the Jellyfish item, based on Duck Game's [RC Controller](https://duckgame.fandom.com/wiki/RC_Controller).

More tweaks may come in follow-up PRs

## Notable Changes from [v1](https://github.com/fishfolk/jumpy/pull/895)

- The camera follows any entity with the new `CameraSubject` component
  - Entity must also have a `Transform` and `KinematicBody` to work properly
- The flappy jellyfish collides with solids
  - The flappy is now much more jerky and overly responsive to user input, I may add acceleration to the left/right movement so that it floats around more to give it a more natural feel
- Control the jellyfish like Duck Game's RC Controller
- The flappy is tied to the jellyfish hat, not the player
  - If the driver dies/drops the jellyfish the flappy doesn't explode, another player can pick it up and take control
- Despawn & dehydrate the jellyfish item when the flappy explodes

## ToDo

- ~Try to get the jellyfish on the player's head~ (will handle in another PR for cosmetic changes)

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -31,8 +31,8 @@ use crate::{prelude::*, settings::PlayerControlMapping};
 pub mod prelude {
     pub use super::{
         attachment::*, audio::*, bullet::*, camera::*, damage::*, debug::*, editor::*, editor::*,
-        elements::prelude::*, globals::*, input::*, item::*, lifetime::*, map::*,
-        map_constructor::*, metadata::*, physics::*, player::*, random::*, utils::*, FPS,
+        elements::prelude::*, flappy_jellyfish::*, globals::*, input::*, item::*, lifetime::*,
+        map::*, map_constructor::*, metadata::*, physics::*, player::*, random::*, utils::*, FPS,
         MAX_PLAYERS,
     };
 }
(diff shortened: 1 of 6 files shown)


zicklag:
FYI I merged your other PR so you'll need to rebase.

Also, when testing this, it seems that you can't control the jellyfish anymore. The jellyfish spawns when you activate the item, but the player is still moving normally.

nelson137:
@zicklag are you holding down the use/fire button? When you let go you lose control.

zicklag:
Ohhhh, I haven't red the wiki page yet. That's actually really cool. And you got collisions working well, it even interacts with sproingers which is fun. :)

In that case, all we need is the rebase.

nelson137:
Earlier I discovered that if you stop driving while the jellyfish is on a spoinger it goes flying up in the air.... I'll need to fix that, I think it's not clamping the y velocity when no one's driving

nelson137:
That's fixed, let me know if you find anything else off!

zicklag:
I think it's good enough to merge!

PR #900, opened 2024-01-13
feat(networking): Re-implement networking from bones + re-add online matches

MaxCWhitehead:
This adds menu + gameplay for online matches back into Jumpy. Some work left to do, but I think this is a good checkpoint, online matches work, lan is still disabled (this is up next).

TODO:
- Lan
- Cleanup menu code a bit more
- Pause menu (right now disabled, need to figure out how to do this without disabling session as I think that causes issues)

### DIFF
--- a/.github/workflows/web-demo.yml
+++ b/.github/workflows/web-demo.yml
@@ -31,7 +31,7 @@ jobs:
       - uses: jetli/wasm-bindgen-action@v0.1.0
         name: 🔧 Wasm Bindgen
         with:
-          version: "0.2.89"
+          version: "0.2.90"
 
       - uses: actions/cache@v3
         name: ♻️ Cache Cargo
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1119,7 +1119,7 @@ dependencies = [
 [[package]]
 name = "bones_asset"
 version = "0.3.0"
-source = "git+https://github.com/fishfolk/bones#06d5efcace0f07986735fe872590449a85ca2104"
+source = "git+https://github.com/fishfolk/bones#d560443794970526e72c848b4c1ade71ccf4de6a"
 dependencies = [
  "anyhow",
  "append-only-vec",
@@ -1150,7 +1150,7 @@ dependencies = [
 [[package]]
 name = "bones_bevy_renderer"
 version = "0.3.0"
-source = "git+https://github.com/fishfolk/bones#06d5efcace0f07986735fe872590449a85ca2104"
+source = "git+https://github.com/fishfolk/bones#d560443794970526e72c848b4c1ade71ccf4de6a"
 dependencies = [
  "anyhow",
  "bevy",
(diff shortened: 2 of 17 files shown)


zicklag:
@MaxCWhitehead this is such a huge help, thank you! I pushed a few commits to fix ultra minor things, and there are still a couple bugs I found:

- In offline play, if you try to pick up an item, and you hold down the pick up button, it will pick it up and drop it over and over super fast, like it's not realizing that you are still holding the button and it things you are pressing the button over and over.
- In online play, it seems like holding down buttons doesn't work. At least, if hold down the jump button, the player doesn't float down slowly.

Also, in the player selection screen on online play you have to press the jump button once before you can select your player which is a little unintuitive.

MaxCWhitehead:
@zicklag Yeah I couldn't remember how the pressing input to select player thing worked - I thought maybe this behavior was intended to select the control source (press space or a on controller), but can revisit that.

Good catch on the input bugs, shouldn't be too hard to track down. Will look into that.

MaxCWhitehead:
I had the changes prepped so thought I'd toss them in here for lan: Depends on this PR (https://github.com/fishfolk/bones/pull/308) probably will fail CI until that is merged + bones version updated here.

I haven't gotten lan service discovery to work locally (though I am unsure if I should expect it to work with two clients on same host) - will try to test this with a second machine later.

Anywho will go fix the other bugs.

zicklag:
> I thought maybe this behavior was intended to select the control source (press space or a on controller), but can revisit that.

Oh, that's a good point. :thinking: When I first got into it I couldn't figure out why I couldn't select my character, but maybe the workflow does make sense and we just need a way to indicate that you have to press a button to select your controller.

Maybe just leave it as is, and we can see about improving the UX afterwards.

> I had the changes prepped so thought I'd toss them in here for lan: Depends on this PR (https://github.com/fishfolk/bones/pull/308) probably will fail CI until that is merged + bones version updated here.

Cool, I queued the merge of that bones PR.

> I haven't gotten lan service discovery to work locally (though I am unsure if I should expect it to work with two clients on same host) - will try to test this with a second machine later.

It should work with two games on the same host, though I think it didn't work for somebody on some version of Linux, because it wasn't enabled or something. I'm not 100% sure how out-of-the-box configured mDNS is on different platforms, but I _think_ it should be pretty standard.

MaxCWhitehead:
Fixed the input jittering bug in local (forgot to call one of the trait functions for saving prev frame input).

We were replicating just_pressed for jump in networking  - I think we were doing this post refactor but might be wrong, either way changing this to jump_pressed fixed it.

For player select, you now only have to press confirm once, and I updated the text to show all possible bindings for confirm. The bindings confirm should now be dynamic as well in local match (when someone takes a control source it is removed from other slots).

zicklag:
Tested this out and I think it's merge ready! This is huge to get back in!

PR #901, opened 2024-01-14
feat: separate alarm emote from face animations

### LINKED ISSUE #803: Split Face Animation From Exclamation Mark Animation
Right now the exclamation marks in the fish emotes are attached to the facial animations. This should be split up into two separate sprites so that that players that don't have customized facial expressions yet can still have the exclamation mark as part of the emote.

nelson137:
- Closes #803 

## Changes

- Remove the alarm exclamation point from player face skins
- Create a new player asset type (`assets/player/emotes/alarm.emote.yaml`)
  - The emote sprite is the same size as the face sprite and has the same offset. This made it easy to make sure it's in the same place relative to the player, but it can be shrunk to just the emote region if necessary.


zicklag (approved):
Great, thanks!

PR #902, opened 2024-01-15
fix: kick bomb randomness

### LINKED ISSUE #720: Kick Bomb Bounce Is Unpredictable
@erlend-sh What is the logic supposed to be as far as when the kick bomb should bounce off of the player and when it should kill them.

Right now the behavior seems kind of hard to predtict ( I'm not sure why yet ) but I don't know how it's supposed to be.

nelson137:
Closes #720

## Changes

- Shorter fuse time: 4s, same as grenade
- Less bouncy
- Explodes on the third kick from a player

### DIFF
--- a/assets/elements/item/kick_bomb/kick_bomb.yaml
+++ b/assets/elements/item/kick_bomb/kick_bomb.yaml
@@ -1,5 +1,5 @@
 fuse_time: 8s
-kick_velocity: [10, 6]
+kick_velocity: [10, 2]
 throw_velocity: 10
 damage_region_size: [60, 60]
 damage_region_lifetime: 0.6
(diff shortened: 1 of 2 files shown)


nelson137:
I think this feature is ready for review besides maybe some tweaks to the physics. And since I'm working in this item I want to double check that some of the behavior is still what we want:

https://github.com/fishfolk/jumpy/issues/720#issuecomment-1890843359 (linking this here for continuity)

The bomb will explode on collision with a sliding player or a player's back. I'm not sure if that second part was done because only bouncing off the front of a player makes the logic simpler, but if that's no longer desired I can fix that here too :)

zicklag:
I think this is good! I'm pretty sure it's supposed to explode off of the back of a player, since they can't reasonably kick it when it's behind them ( maybe 🤷‍♂️ ), but if we want to change that later we can.

Issue #43, opened 2021-07-27
Special tiles


erlend-sh:
There’s a great number of special tiles to be made:

- Slippery
- Death spikes
- Door
- Mud
- Some equivalent of water
- Some equivalent of lava
- Whirlpool
- ???

olefasting:
We can start thinking about implementing this, now that we have the new map format. I will be implementing `EnvironmentItem` now, which is intended to represent things such as this, as well as sproingers etc.

Zac8668:
Hey! I will claim this, I will use the `EnviromentItem` as olefasting quoted on the comment above, on a similar manner to the sproinger, I'm just not so sure how the Lava and Water equivalents would work, maybe it could be a tile and not an `EnviromentItem`, that would open more possibilities when making a map imo.

olefasting:
> Hey! I will claim this, I will use the `EnviromentItem` as olefasting quoted on the comment above, on a similar manner to the sproinger, I'm just not so sure how the Lava and Water equivalents would work, maybe it could be a tile and not an `EnviromentItem`, that would open more possibilities when making a map imo.

The `EnvironmentItem` has been removed from the API and these things are now implemented explicitly. See the sproinger for an example..

PR #903, opened 2024-01-16
feat: non-tile solids

nelson137:
Implement collisions for non-tile solids and a way to define them for elements in a map's YAML. This makes the `CollisionWorld` account for entities with the `Solid` component.

I started out trying to implement a door for #43 but quickly realized I couldn't make a closed door solid. This is its own PR because I want to make sure I'm going in the right direction here and discuss the implementation.

https://github.com/fishfolk/jumpy/assets/25290530/3851d2b3-8f58-41c0-8462-9b641334d63a

### Defining solids

Elements can be defined in each layer of a map's YAML file. Solids can now be defined on an element with a `solid` object.

Example element in a map layer:

```yaml
  - pos: [630.0, 365.0]
    element: core:/elements/environment/demo_solid_box/demo_solid_box.element.yaml
```

Now, with a solid:

```yaml
  - pos: [630.0, 365.0]
    element: core:/elements/environment/demo_solid_box/demo_solid_box.element.yaml
    solid:
      enabled: true
      pos: [630.0, 365.0]
      size: [16.0, 16.0]
```

The solid has its own position and size since a collider may need to be separate from the sprite. For example, the door sprite will need to be very wide for when it's open but the collider for a closed door will be very thin.

This was done by adding the new struct `ElementSolidMeta` to `ElementSpawn`. I wanted it to be a `Maybe<ElementSolidMeta>` but got an error from bones. My next best option was letting it be default-initialized for all elements but provide an `enabled` boolean that must be `true` for the solid to be created. Not ideal but it works.

Another downside is that the size of the solid collider has to be defined *in the map*, whereas the size of the `KinematicBody` is often defined in the element's `.yaml` data file and the size of the sprite is defined in the element's `.atlast.yaml`. Maybe this can eventually be moved into an element's asset configs.

### Collision detection

Solids get a `Collider` component like kinematic bodies, which are also synced to the rapier colliders. This allows game code to easily control the collider. e.g. the door will want to disable the collider when it's open, which can be done by setting `Collider.disabled` to `true`.

I'm not convinced my changes to `CollisionWorld::move_{horizontal,vertical}` are correct, but this seems to mostly work.

Since the `CollisionWorld::tile_collision{,_filtered}` methods now detect *any* collision they should be renamed to `CollisionWorld::collision{,_filtered}`. If this is looking good so far I will make that change.

### Game behavior

I added a few demo boxes to dev level 1 for testing. It works great from the testing I did except for a couple minor bugs. If you slide into the box to the left of where you spawn you stop short of the box. But if you walk up to the box you collide with it as expected. Sliding into the one in the far bottom right of the map behaves like you would expect. This is all shown in the video.

Additionally, critters cause a lot of collision warnings:

> 2024-01-16T22:02:52.902798Z  WARN jumpy::core::physics: Collision test error resulting in physics body stuck in wall at Rect { min: Vec2(712.0, 96.1), max: Vec2(722.0, 106.1) }

Snails walk straight through solids, causing these warnings. Crabs may attempt and fail to surface under a solid after burrowing, also causing these warnings. I'm not sure what to do about this yet.

### DIFF
--- a/src/core/elements.rs
+++ b/src/core/elements.rs
@@ -94,6 +94,10 @@ pub struct DehydrateOutOfBounds(pub Entity);
 #[repr(C)]
 pub struct ElementHandle(pub Handle<ElementMeta>);
 
+#[derive(Clone, HasSchema, Default, Deref, DerefMut)]
+#[repr(C)]
+pub struct ElementSolid(pub Entity);
+
 #[derive(Clone, HasSchema)]
 #[schema(no_default)]
 pub struct ElementKillCallback {
--- a/src/core/elements.rs
+++ b/src/core/elements.rs
@@ -45,12 +45,11 @@ pub struct ElementMeta {
     pub plugin: Handle<LuaPlugin>,
 }
 
-#[derive(HasSchema, Default, Clone, Debug)]
+#[derive(HasSchema, Default, Debug, Clone, Copy)]
 #[type_data(metadata_asset("solid"))]
 #[repr(C)]
 pub struct ElementSolidMeta {
-    pub enabled: bool,
-    pub pos: Vec2,
+    pub offset: Vec2,
     pub size: Vec2,
 }
 
--- a/src/core/elements.rs
+++ b/src/core/elements.rs
@@ -49,6 +49,7 @@ pub struct ElementMeta {
 #[type_data(metadata_asset("solid"))]
 #[repr(C)]
 pub struct ElementSolidMeta {
+    pub disabled: bool,
     pub offset: Vec2,
     pub size: Vec2,
 }
--- a/src/core/elements.rs
+++ b/src/core/elements.rs
@@ -45,6 +45,15 @@ pub struct ElementMeta {
     pub plugin: Handle<LuaPlugin>,
 }
 
+#[derive(HasSchema, Default, Clone, Debug)]
+#[type_data(metadata_asset("solid"))]
+#[repr(C)]
+pub struct ElementSolidMeta {
+    pub enabled: bool,
+    pub pos: Vec2,
+    pub size: Vec2,
+}
+
 #[derive(HasSchema, Deserialize, Clone, Debug)]
 #[repr(C)]
 pub struct ElementEditorMeta {
--- a/src/core/map.rs
+++ b/src/core/map.rs
@@ -162,9 +162,6 @@ fn spawn_map(
     mut camera_states: CompMut<CameraState>,
     mut spawned_map_layer_metas: CompMut<SpawnedMapLayerMeta>,
     mut spawned_map_meta: ResMutInit<SpawnedMapMeta>,
-    mut element_solids: CompMut<ElementSolid>,
-    mut solids: CompMut<Solid>,
-    mut colliders: CompMut<Collider>,
 ) {
     if map_spawned.0 {
         return;
@@ -271,27 +268,6 @@ fn spawn_map(
                 Transform::from_translation(element_meta.pos.extend(layer_z)),
             );
             element_handles.insert(element_ent, ElementHandle(element_meta.element));
-
-            if element_meta.solid.enabled {
-                let solid_ent = entities.create();
-                element_solids.insert(element_ent, ElementSolid(solid_ent));
-                solids.insert(
-                    solid_ent,
-                    Solid {
-                        pos: element_meta.solid.pos,
-                        size: element_meta.solid.size,
-                    },
-                );
-                colliders.insert(
-                    solid_ent,
-                    Collider {
-                        shape: ColliderShape::Rectangle {
-                            size: element_meta.solid.size,
-                        },
-                        ..default()
-                    },
-                );
-            }
         }
     }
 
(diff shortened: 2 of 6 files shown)


zicklag:
Cool, thanks for checking this out! I'm going to do a proper review of this as soon as I can, probably over the week-end if I can't find time at the end of the next couple days.

nelson137:
Great! No rush, I already added 1 thing I missed, and now I think I need to support multiple solids per element for the door 😅

src/core/elements/demo_solid_box.rs
```diff
@@ -0,0 +1,45 @@
+use crate::prelude::*;
+
+#[derive(HasSchema, Default, Debug, Clone)]
+#[type_data(metadata_asset("demo_solid_box"))]
+#[repr(C)]
+pub struct DemoSolidBoxMeta {
+    pub atlas: Handle<Atlas>,
```
zicklag, line 7: Regarding the way that we define solids, I'm pretty sure we don't want to add the new element metadata to the map. Instead we should put all of the metadata for the solids in whatever kind of elements that we want to spawn.

For instance, now that we have the physics implemented, any element, such as a "door" element, can spawn it's own entities with a `Solid` component to serve as a collision, while also spawning whatever sprites it needs with different sizes, etc.

So all of the spawning logic would be implemented by the door, instead of putting any of that info in the map metadata.

Does that make sense for what you were thinking?

zicklag (commented):
>  If you slide into the box to the left of where you spawn you stop short of the box.

That's actually a longstanding bug, not introduced here. I think it has something to do with how we set the position of the player when he "stands up" from sliding. I never figured out exactly why it happened or why it's different on the left than it is on the right side of the player.

nelson137:
> That's actually a longstanding bug, not introduced here

I can open an issue for that and then pick it up sometime after this

> Instead we should put all of the metadata for the solids in whatever kind of elements that we want to spawn...

That makes a lot of sense, and I like that way better than putting solids in the map.

nelson137:
I've finished making those changes.

I also realized that I was only using `Collider` for its `rapier_handle` and `disabled`. I decided to move all of this functionality into `Solid`, so they no longer need a `Collider`. I don't believe using a `Collider` for these kinds of entities is appropriate since it seems tailored for things that have more normal physics (gravity, friction, velocity, interaction with jump-through platforms like wood, etc.).

Another option would've been to go full `KinematicBody` + `Collider` and make `Solid` just a marker component with no data. But IMO solids are fundamentally different kinds of entities whose only overlap with other phsyics objects is collision detection, and therefore they merit their own special treatment. Let me know if you disagree / have any thoughts on this.

This change means that a `Solid` has full control over the rapier collider, making it a nice interface to update the size, position, and disabled of the collider in-game. I updated the demo to show this.

Anyway, if everything looks good I can drop those DELETEME commits.

src/core/debug.rs
```diff
@@ -23,6 +23,16 @@ pub struct DebugSettings {
     pub show_pathfinding_lines: bool,
 }
 
+impl Default for DebugSettings {
+    fn default() -> Self {
+        Self {
+            show_kinematic_colliders: true,
```
zicklag, line 29: This should be set back to false before merging.

zicklag (changes requested):
This looks good, other than hiding the collision shapes.

There is one other but I noticed, but I don't know how to reproduce. When I started the game the first time, for some reason the collision shape of the demo box on the right started moving and ended up stretched out like the red annotated box in this screenshot:

image

And it was moving, but not the sprite, just the collision shape that you could see because the debug lines! I'm not sure what would cause that. I restarted the match and it didn't happen again, but it's something to keep an eye out for I guess if it happens later or we can figure out what causes it.

nelson137:
> And it was moving, but not the sprite, just the collision shape

Sorry, my last message wasn't very clear. I did that on purpose to show that changing the properties of `Solid` syncs with the collider. So for that one box I change the position, size, and toggle disabled every second.

And the change to `DebugSettings` should be in one of the DELETEME commits, and I will drop those before rebasing

src/core/elements/demo_solid_box.rs
```diff
@@ -0,0 +1,96 @@
...
+    if let Some((_, (_, solid_ent))) = entities
+        .iter_with((&boxes, &element_solids))
+        .find(|(_, (b, _))| b.0 == 0)
+    {
+        let solid = solids.get_mut(**solid_ent).unwrap();
+        solid.pos.x += 0.05;
+        solid.size.x += 0.03;
+        solid.disabled = time.elapsed_seconds() as u32 % 2 == 0;
+    }
+}
```
nelson137, line 96: This controls the collider of that one box

zicklag:
Oh, OK, you're right, I forgot about that in your comment. Cool, that's actually great! So then I'm all good with this, you can drop those commits and I'll merge it!

src/core/elements/demo_solid_box.rs
```diff
@@ -0,0 +1,96 @@
...
+    time: Res<Time>,
+) {
+    if let Some((_, (_, solid_ent))) = entities
+        .iter_with((&boxes, &element_solids))
+        .find(|(_, (b, _))| b.0 == 0)
+    {
+        let solid = solids.get_mut(**solid_ent).unwrap();
+        solid.pos.x += 0.05;
+        solid.size.x += 0.03;
+        solid.disabled = time.elapsed_seconds() as u32 % 2 == 0;
```
nelson137, line 94: This controls the collider of that one box

zicklag (approved):

PR #908, opened 2024-01-20
feat(profiler): Re-add puffin profiler (F10)

MaxCWhitehead:
Re add puffin profiler. Haven't added any new scopes / fleshed anything out, but got it handy at least for now.


src/profiler.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/// Designate session that marks profiler's frame boundary.
+///
+/// Installing to game session is recommended, however this will not work if
+/// trying to profile outside of match or while game session is paused.
+pub fn install_frame_marker(session: &mut Session) {
+    session
+        .stages
+        .add_system_to_stage(CoreStage::First, mark_new_frame);
```
zicklag, line 19: We should probably install the `mark_new_frame` system as a [game system](https://fishfolk.github.io/bones/rustdoc/bones_lib/struct.Game.html#structfield.systems) so that it runs before any created sessions.

zicklag (changes requested):
Awesome, puffin is pleasantly easy.

zicklag (approved):

PR #909, opened 2024-01-23
feat: add audio settings

nelson137:
Add an "Audio" page to settings with volume controls. It features a main volume slider and sliders to control music and sound effects separately.

This replaces `AudioEvents` with a new `AudioCenter` that manages all audios (I would've liked to use `AudioManager` but that's [taken by bones](https://github.com/fishfolk/bones/blob/091c97883406be8a9ad2bb8683a114096afae23b/framework_crates/bones_framework/src/render/audio.rs#L26)).

before:

```rust
audio_events.play(*explosion_sound, *explosion_volume);
```

after:

```rust
audio_center.play_sound(*explosion_sound, *explosion_volume);
```

For sounds, the volume is calculated as `main_volume * effects_volume * requested_volume` where the requested volume is the argument to `play_sound`. A similar method exists for music.

Sound effects are modeled by entities with an `Audio` component. When they are stopped/finish the entity is killed. The music system only supported 1 song at a time, so music is stored on the `AudioCenter` as an `Option<Audio>`. This can easily be refactored to use entities, like sound effects do, in the future if needed.


nelson137:
I just upgraded shlex (v1.2.0) to v1.3.0 to fix the failing task due to the security advisory

zicklag (approved):
Love it! This is great, thanks!

PR #911, opened 2024-01-24
fix: sliding collisions

### LINKED ISSUE #910: bug: incorrect collision when player slides into the right side of a solid
### Description

When a player slides into the right side of a solid object they are stopped just before it, leaving a gap.

https://github.com/fishfolk/jumpy/assets/25290530/3851d2b3-8f58-41c0-8462-9b641334d63a

### To Reproduce

1. Walk left toward a solid object
2. Start sliding

### Expected Behavior

Player stops up against the collider -- there should be no gap when sliding into an object.

### Additional Context

This bug is not present for the left sides of objects.

### Log Messages

_No response_

nelson137:
Closes #910

This was more simple than I expected. When a sliding player collides with something and stops moving, the transform adjustment to stand up didn't take into account the direction the player is facing.

https://github.com/fishfolk/jumpy/assets/25290530/aecf7cf4-72ef-427b-8f16-73000f43ad5f

### DIFF
--- a/src/core/elements/slippery_seaweed.rs
+++ b/src/core/elements/slippery_seaweed.rs
@@ -88,7 +88,7 @@ pub fn update(
                 .actor_collisions(p_ent)
                 .contains(&seaweed_ent)
             {
-                state.current = "core::incapacitated".into();
+                state.current = *incapacitated::ID;
                 continue;
             }
         }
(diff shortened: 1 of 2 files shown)


nelson137:
This advisory check task will pass once #909 is merged and I rebase

zicklag:
Oh, sweet! That's so nice to have fixed. It's one of those small wrinkles that has bugged me since I ported the original physics. :)

PR #915, opened 2024-02-03
character select screen tweak

nickrehac:
make character/hat selection more obvious by keeping eyes mostly in the center of the character panel

Screenshot from 2024-02-03 17-02-54
Screenshot from 2024-02-03 17-02-47
Screenshot from 2024-02-03 17-02-42

### DIFF
--- a/src/ui/main_menu/player_select.rs
+++ b/src/ui/main_menu/player_select.rs
@@ -440,28 +440,6 @@ fn player_select_panel(
                     direction.x > 0.0,
                     &state.hats,
                     &mut slot.selected_hat);
-                    
-                /*
-                let current_hat_handle_idx = state
-                    .hats
-                    .iter()
-                    .enumerate()
-                    .find(|(_, handle)| **handle == slot.selected_hat)
-                    .map(|(i, _)| i)
-                    .unwrap_or(0);
-
-                let next_idx = if direction.x > 0.0 {
-                    (current_hat_handle_idx + 1) % state.hats.len()
-                } else {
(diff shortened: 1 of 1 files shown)


zicklag (changes requested):
Hey there, thanks for the PR!

One thing I would like to do is make sure that the player stays the same size before and after selecting the skin. For example, in the screenshot below, the player on the left that hasn't been locked in, is smaller than the skin after you confirm the player selection. 

image

We should try to keep the player the same size before and after confirming.

Another minor tweak is to use the pixel font and size for the hat selection name:

image

I would also like just a little bit more space between the hat's label, and the hat itself. Just a little. 🤏

nickrehac:
Thanks for the feedback!
I looked into the character scaling and found the sprite sheet has a considerable amount of padding which throws off the ui and might be better suited as part of the ui construction itself. I'm going to check if that padding is needed elsewhere

nickrehac:
fixed the size and added spacing.
characters drop down when selected which could be an indicator of selection status, but is actually egui being quirky
thoughts?

Screenshot from 2024-02-06 13-57-11

zicklag:
Hi @nickrehac, sorry for the delay. Life has had me incredibly busy recently. Could you push the code with the updated size and spacing?

As far as the egui issue, there must be some way to fix it, but me or someone else can try to take a look at it if we get the chance. I do want to get that fixed before merging.

nickrehac:
Here you go, and no worries.
I might take a crack at this last bit soon. It seems that calculating the sprite size each frame is unavoidable

PR #917, opened 2024-02-24
build(steam): Steam packaging utility + linux launch scripts

### LINKED ISSUE #904: Steam Deck: Graphics scaled up too big
The bevy_window crate computes scale factor in WindowResolution based on OS reported pixel density. This seems to not end up right on the deck. 

Setting scale_factor_override = 1.0 fixes graphics scaling issues on deck, from my research this is probably less of an issue with bevy's computation of scale factor and more that the deck doesn't report accurate values, but will look into it more.

Worse case scenario we have to detect steamdeck and adjust this value.

**EDIT:** The TLDR of my comments is that I think we should include some special launch shell scripts for linux distribution through steam. If using steam deck display, need to set a variable to fix scale. Doing this in shell / on game launch side keeps the options close to user, and doing this in engine / game code is harder and less accessible to tweak.

MaxCWhitehead:
Add a steam_package.sh script that pulls releases from github, rezips as .zip such that they can be uploaded to steamworks for respective platforms.

For linux, we also install some launch scripts. There's logic here to fix #904. Isn't pretty but seems to work, scaling is correct in steam deck game mode + desktop mode. Should not impact other linux platforms. 

Need to test again with steam deck with external display, but I think this should work based on last I was exploring solutions to #904.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ trace-*.json
 perf.data*
 callgrind.out*
 .cargo
+steam/packaged/**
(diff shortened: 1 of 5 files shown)


zicklag:
Awesome!

@orhun do you have any thoughts on this? It seems pretty simple, but since it was related to builds I wondered if you might have any ideas or anything we should consider.

PR #919, opened 2024-03-02
feat(network-diagnostics): Install bones network debug window

MaxCWhitehead:
Install the bones network debug egui plugin / add a button to open it from debug window.

Companion PR to https://github.com/fishfolk/bones/pull/345

### DIFF
--- a/src/core/physics/collisions.rs
+++ b/src/core/physics/collisions.rs
@@ -301,7 +301,7 @@ impl<'a> CollisionWorld<'a> {
             if !self.entities.is_alive(entity) {
                 // Remove any collisions with the killed entity from the collision cache.
                 let mut collisions = collision_cache.collisions.borrow_mut();
-                let colliding_with = collisions.remove(&entity);
+                let colliding_with = collisions.swap_remove(&entity);
                 if let Some(colliding_with) = colliding_with {
                     for other_entity in colliding_with {
                         if let Some(collisions) = collisions.get_mut(&other_entity) {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -92,9 +92,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.8.7"
+version = "0.8.10"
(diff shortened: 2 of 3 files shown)


zicklag:
I merged the bones PR so you should be able to run `cargo update` to make CI pass for this now.

src/core/physics/collisions.rs
```diff
@@ -301,7 +301,7 @@ impl<'a> CollisionWorld<'a> {
             if !self.entities.is_alive(entity) {
                 // Remove any collisions with the killed entity from the collision cache.
                 let mut collisions = collision_cache.collisions.borrow_mut();
-                let colliding_with = collisions.remove(&entity);
+                let colliding_with = collisions.swap_remove(&entity);
```
zicklag, line 304: What is the effect of this change?

MaxCWhitehead: Ah I should've commented on that - remove was deprecated in index-map crate due to ambiguity. Remove already did a `swap_remove`, they deprecated and require calling `swap_remove` or `shift_remove`, to clarify if user intends to remove and preserve ordering, or swap with last element for performance while modifying ordering.

Behavior has not changed, fixes deprecation warning. See context:

https://github.com/indexmap-rs/indexmap/blob/184fe4bfcba20e21242c924220170b98be157d45/src/map.rs#L645-L653

zicklag (approved):

MaxCWhitehead:
Merged main to resolve cargo.lock conflict

zicklag:
I think there might be a security warning that is going to block merging. Hopefully we can just run `cargo update` to fix it if that's the case.

MaxCWhitehead:
> I think there might be a security warning that is going to block merging. Hopefully we can just run `cargo update` to fix it if that's the case.

I just ran cargo update in the merge from main - so it seems these are still around. I can take a swing at resolving some of these cargo deny errors if you would like.

zicklag:
Oh, I've already got the advisories set as non-required for merge, so it looks like all we need to do is add `Unicode-3.0` to our allowed licenses list.

MaxCWhitehead:
> Oh, I've already got the advisories set as non-required for merge, so it looks like all we need to do is add `Unicode-3.0` to our allowed licenses list.

PR opened: https://github.com/fishfolk/jumpy/pull/923

PR #926, opened 2024-03-21
fix(devpack): Update devpack version to 0.10.1 to match game version

MaxCWhitehead:


zicklag (approved):
If you're ever waiting on review on something minor and it helps you to just get it in, feel free to merge these yourself, too. I don't want to slow you down at all if it helps. 👍️

PR #927, opened 2024-03-21
feat(physics): Support Dynamic bodies

MaxCWhitehead:
This is the base of core changes to support ragdolls, allows us to simulate dynamic bodies in rapier.

DynamicBody component may be added to an entity that has KinematicBody, and if enabled, body will simulate physics. Transitioning between kinematic / dynamic and adjusting rapier data + collider is handled on next physics update.

This change should have no impact on existing kinematics, scene queries, etc. Usage of these changes will come in other PR soon.

**Switched to fork of rapier (temporarily hopefully):**
I switched us to a fork of rapier to fix a bug in which changing a rapier rigidbody from kinematic to dynamic will not put it in active dynamic set, so gravity is not applied. It seems they rely on contact graph to add body to this set, but if not in contact with anything it breaks. I have a PR open for a fix in rapier.
- I also updated version, slight change in debug draw code is because of changes there.

**Some notes on changes:**
- Some helper functions are included for pushing simulation commands for dynamic body that apply changes to rapier body before next physics step.
- There is a mutate_rigidbody function on CollisionWorld that makes accessing rapier body for immediate mutation slightly more convenient.
- Previously bodies were created as dynamic (but we used them as kinematic anyway), now they are actually flagged as kinematic in rapier. This does not impact collision events at all.

There's a decent chunk of physics stuff in collision module now, at some point maybe should detangle this, but probably fine for now. I renamed some functions to make things more clear.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3602,17 +3602,6 @@ checksum = "51d515d32fb182ee37cda2ccdcb92950d6a3c2893aa280e540671c2cd0f3b1d9"
 
 [[package]]
 name = "num-derive"
-version = "0.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "876a53fff98e03a936a674b29568b0e605f06b29372c2489ff4de23f1949743d"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 1.0.109",
-]
-
-[[package]]
-name = "num-derive"
 version = "0.4.2"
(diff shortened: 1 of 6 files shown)


zicklag (approved):
Nice. I like how small this diff is. 🚀

MaxCWhitehead:
> Nice. I like how small this diff is. 🚀

There's more haha but going in logical chunks at least so it's not a massive change soup :)

PR #928, opened 2024-03-22
feat(physics): Contact filtering + changing shape + jump through for simulated bodies

MaxCWhitehead:
Another chunk of features implemented to support ragdolls + dynamic bodies in Jumpy.

### **Contact filtering:**
Add bit flags for rapier `CollisionGroup` and `SolverGroup`.
- The Collision group filters both contact forces and events, these are configured such that there is no change here in collision events.
- Solver group then filters contact forces, this is used to make sure dynamic actors only simulate collision with tiles (solid + jump through) and solids, not other kinematic (or dynamic) actors.

### **Jump Through for Dynamics:**
Dynamics collide with jump through tiles. A contact modification physics hook is implemented so dynamics moving upward through jump throughs have contact thrown out, and kept when falling.
- Only the contact filtering `PhysicsHook` is used, and is only enable on dynamic bodies while they are simulating.

### **Changing actor shape:**
I found that using a capsule works better for player ragdoll then rectangle, which requires being able to change the shape of a collider. A helper function is added to do this correctly, changing shape is only currently supported for actors. 

### **Handling `Stop` collision events for removed colliders:**
Changing an actor's shape involves removing collider and re-adding it. This triggers a stop event from rapier. These events provide collider handles, yet we map this to entities from userdata. The consequence of this is that when the stop event is dispatched, the collider handle is already invalid (generation incremented), meaning we no longer can access user data to determine what entity should be removed.
- We now have a function that should be called on removal of collider in `CollisionCache` that temporarily caches the handle + entity, which is used as fallback when receiving stop event if collider handle is not valid.

This is not a great system, if someone removes a collider and does not call this, our collision event will not be removed causing bugs. Might be able to make this better, but for now will just have to make sure if we get fancy removing a collider (which is already a delicate / involved operation) this is called. The function to change actor shape calls this for us.

### DIFF
--- a/src/core/physics/collisions.rs
+++ b/src/core/physics/collisions.rs
@@ -28,6 +28,7 @@ pub struct RapierContext {
     pub rigid_body_set: rapier::RigidBodySet,
     pub collider_shape_cache: ColliderShapeCache,
     pub collision_cache: CollisionCache,
+    pub physics_hooks: PhysicsHooks,
     pub physics_pipeline: rapier::PhysicsPipeline,
     pub islands: rapier::IslandManager,
     pub impulse_joints: rapier::ImpulseJointSet,
@@ -49,6 +50,7 @@ impl Clone for RapierContext {
             rigid_body_set: self.rigid_body_set.clone(),
             collider_shape_cache: self.collider_shape_cache.clone(),
             collision_cache: self.collision_cache.clone(),
+            physics_hooks: self.physics_hooks.clone(),
             // Probably should keep this around, it's safe to drop data as only temp buffers,
             // but re-creating may hurt performance.
             physics_pipeline: rapier::PhysicsPipeline::default(),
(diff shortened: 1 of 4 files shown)


src/core/physics/collisions.rs
```diff
@@ -182,6 +219,62 @@ pub enum PhysicsError {
...
+            .intersects(SolverGroup::JUMP_THROUGH.bits().into())
+        {
+            jump_through_body = Some(body2);
+            other_body = Some(body1);
+        }
+
+        if jump_through_body.is_some() {
+            let other_body = other_body.unwrap();
+
+            if other_body.linvel().y > 0.0 {
```
MaxCWhitehead, line 271: This isn't quite correct, as if body is already inside jump through and starts falling, it will start colliding and pop downwards a bit faster than expected, but not too worried about that yet.

zicklag:
Sounds good!

PR #931, opened 2024-03-23
feat(physics): Add extra colliders to jump_through tiles for collision with dynamics

MaxCWhitehead:
define some collision info for jump through wood tiles, used to create accurate colliders for collision with ragdoll / dynamics.

The standard tile is still there for all intents and purposes. If there is an extra collider defined, this collider is also added, and is set to solver group such that dynamics will collider with it.
- This additional collider is not used for collision events.
- if additional collider is present, original tile collider no longer has solver group for collisions.

This fixes dynamics colliding with the invisible chunk on thin tiles such as jump through wood.

I think we're done with the trickle of auxiliary features and next PR should be ragdolls 😆

### DIFF
--- a/assets/map/resources/ship_decorations.atlas.yaml
+++ b/assets/map/resources/ship_decorations.atlas.yaml
@@ -2,3 +2,25 @@ image: ./ship_decorations.png
 tile_size: [32, 32]
 columns: 11
 rows: 5
+tile_collision:
+  41:
+    min:
+    - 0.2
+    - 0.75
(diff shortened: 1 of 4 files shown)


MaxCWhitehead:
looks like I messed up PR trying to merge main to fix cargo.lock - will fix

MaxCWhitehead:
> looks like I messed up PR trying to merge main to fix cargo.lock - will fix

Ok fixed

PR #934, opened 2024-03-23
fix: reset hat when dropped

### LINKED ISSUE #930: If player drops hat while translucent during invincibility, hat remains translucent
### Description

If player drops hat while translucent during invincibility, hat remains translucent.

### To Reproduce

1. Spawn into game with hat
2. Press F to ragdoll self, dropping hat (this ragdoll feature is not implemented in this moment, PR coming soon, probably will be by the time someone is fixing this 😸 )
3. Observe hat is translucent and stays that way

### Expected Behavior

Hat should drop without being translucent.

### Additional Context

_No response_

### Log Messages

_No response_

nelson137:
Closes #930 

- Adds a `PlayerCommand` to drop a player's hat and reset its properties
- Replaces all instances of manually dropping a hat to use the new convenience command

### DIFF
--- a/src/core/player.rs
+++ b/src/core/player.rs
@@ -247,6 +247,7 @@ impl PlayerCommand {
         })
         .system()
     }
+
     /// Despawn a player.
     ///
     /// > **Note:** This is different than the [`kill`][Self::kill] event in that it immediately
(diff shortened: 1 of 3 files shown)


MaxCWhitehead:
haha awesome thanks, I literally thought about helper command and doing this and then chose the lazy path. Great minds :) 

Looks good to me! Thx for alpha fix too

PR #935, opened 2024-03-23
fix(networking): Specify local input delay and max prediction window from metadata

MaxCWhitehead:
I opened a PR in bones to expose these settings for networking, instead of hard coding in engine. (https://github.com/fishfolk/bones/pull/354)

The defaults of these values were also changed in bones such that things are feeling much better in network play with ping around 100ms-150ms.

Now we may tune in metadata, though settings only take effect if new network match is created. This will not pass CI until bones PR is merged + bones version updated.

I think retuning this is a quick and big improvement for high-latency network play. I have WIP for visually smoothing corrections - but I think addressing core issues with networking and getting it as good as possible is priority vs trying to improve how corrections look. Will get that shipped at some point, but want to do some playtesting with more people than myself with various net conditions before spending any more time on net corrections, as it's hard to get right.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1006,7 +1006,7 @@ dependencies = [
  "bitflags 2.4.2",
  "cexpr",
  "clang-sys",
- "itertools 0.12.1",
+ "itertools",
  "lazy_static",
  "lazycell",
  "proc-macro2",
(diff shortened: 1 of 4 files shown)


zicklag:
Merged the bones PR, so you can update this, and feel free to merge once it's update.👍️

MaxCWhitehead:
lol apparently cfg! is not compile time or something - have to fix wasm build, gotta run but will fix + merge later.

zicklag:
Weird, `cfg!` should evaluate at compile time to a constant boolean.

MaxCWhitehead:
> Weird, `cfg!` should evaluate at compile time to a constant boolean.

I think this does happen, and it's probably optimized out - but I guess because the code is still included it gives an error. Have to use the attribute I guess.

PR #953, opened 2024-03-30
feat: port periscope

### LINKED ISSUE #151: [SPEC] Periscope Rifle
Periscope Rifle

DRuppFv:
Periscope ported, this closes #151.

It's similar to the musket and is mainly useful if the enemy is higher than you. Used the artwork from #151 but the concept art confused me a bit as it says to be "similar to the musket" but in the art it looks like more a shotgun as it has some bullet spread, but adding a very low bullet spread would be a nice idea here so let me know if you think it should be used.

As it's very similar to the musket, it had a certain amount of code duplication, which I think we could discuss together and come to a decision whether to generalize something elsewhere or whether we want to make adjustments to add more uniqueness to the designs.

shooting_periscope

periscope


erlend-sh:
> As it's very similar to the musket, it had a certain amount of code duplication

That could also make it a good candidate for a Lua implementation. This PR is good as-is, just saying putting a note in for later.

MaxCWhitehead:
Awesome!

PR #954, opened 2024-04-03
feat: port machine gun

### LINKED ISSUE #573: Re-implement Machine Gun

DRuppFv:
This PR closes #573.

I added `bullet_spread` system to the machine gun because I think it balances the game and looks good, and added `empty_cooldown` because it would be really boring if `gun_empty.ogg` kept playing at the same frequency as the normal cooldown, so in this gun we have the shooting cooldown and the empty sound cooldown. I also made a very slight change in the main system for using items because it had to be able to shoot constantly, so now you can keep shoot input pressed with any weapon.

Something I think is important to mention is that the shooting animation is synchronized to the rate of fire, so if you change the rate of fire the animation will also change. It could be easily changed with a new variable in the .yaml so let me know whether you think it should be faster or slower.

The machine gun has two states (idle and shooting) and its only function is to reset and start animations.


assets/elements/item/machine_gun/machine_gun.yaml
```diff
@@ -0,0 +1,21 @@
...
+empty_shoot_sound: ./shoot/gun_empty.ogg
+
+bounciness: 0.3
+can_rotate: true
+body_size: [32, 8]
+fin_anim: grab_2
+angular_velocity: 0.1
+throw_velocity: 360
+grab_offset: [13, -3]
+kickback: 0
```
MaxCWhitehead, line 21: Wondering if it might be fun to try adding in a bit of kickback so player is moved while holding down machine gun?

src/core/elements/machine_gun.rs
```diff
@@ -0,0 +1,305 @@
...
+
+                    let player_sprite = sprites.get_mut(player).unwrap();
+                    let player_flip_x = player_sprite.flip_x;
+                    let player_body = bodies.get_mut(player).unwrap();
+
+                    //Set kickback
+                    player_body.velocity.x = if player_flip_x { 1.0 } else { -1.0 } * kickback;
+
+                    let mut shoot_animation_transform = *transforms.get(entity).unwrap();
+                    shoot_animation_transform.translation.z += 1.0;
```
MaxCWhitehead, line 247: Not worried about merging as this currently is - but I think when possible it's generally preferable to have values like this live inside the metadata yaml, that way they can easily be found and updated, like if the art for weapon changed.

Maybe could add something like `bullet_spawn_offset` to `machine_gun.yaml`?

Cargo.lock
MaxCWhitehead, line 1: It seems a change to Cargo.lock was included in this PR, which is creating a merge conflict with main.

Typically it isn't necessary to update lockfile unless specifically need to / changed dependencies in Cargo.toml.

Could you please adjust to use the Cargo.lock from main so no diff? We updated some packages in main and want to make sure don't undo those changes. If you need any help with this please let me know.

MaxCWhitehead (changes requested):

assets/elements/item/machine_gun/machine_gun.yaml
DRuppFv: That's a good idea! - Looked for a value that doesn't affect when it's a single shot, but does affect if the machine gun is held.

src/core/elements/machine_gun.rs
DRuppFv: Yeah I've thought about that! It would be really important now that we are porting many different weapons, which have very different values due to their size and shape.

I guess I could make a separate PR for that since it would be a general update?

Cargo.lock
DRuppFv: I'm not sure about what happened here. I probably accidentally executed a general cargo update command. Just ran `git restore --source main -- Cargo.lock`, hope that's enough.

src/core/elements/machine_gun.rs
MaxCWhitehead: Up to you - if you'd like to follow up in another PR, that sounds good to me.

Don't necessarily need to worry about doing this for all existing weapons either (not sure if that's what you had in mind). But if you want to - feel free :)

MaxCWhitehead:
@DRuppFv It looks like the continuous integration found some issues with the formatting in `machine_gun.rs`, could you run `cargo fmt` and update this?

Thank you!

MaxCWhitehead (approved):
Thank you very much for your work here @DRuppFv  :)

PR #959, opened 2024-04-04
feat: bullet spawn offset in yaml

DRuppFv:
I did what @MaxCWhitehead told me earlier, created a `bullet_spawn_offset` field in `.yaml`. I added this to all the weapons because it makes it easier for new contributors to locate and modify in case we change the artworks.


MaxCWhitehead (approved):
Looks great thank you!

PR #991, opened 2024-04-25
feat(input): Allow movement / menu nav with dpad and joystick

MaxCWhitehead:
This adds `movement_alt` control bindings that default to dpad on gamepad and None on keyboard mappings.

This allows the dpad or joystick to be used to move player and navigate menu. We may not want this if we choose to use dpad for other stuff later, but for now I think this is nice as it avoids player having to remap gamepad controls to use dpad. Menu navigation on joystick is rough as have to flick the joystick repeatedly, dpad is a better experience for me / this came up in feedback too.

Added a helper func to merge both the joystick and dpad input - the `movement` binding (joystick) is preferred, but if no input it then takes the `movement_alt` (dpad).

Had a rough time testing this as can't get xbox pro controller 2 to work (probably macOS issue as it's got bad driver support), or gilrs issue. And my PS5 controller crashes (is a gilrs issue) lol. but I tested on steamdeck and this seems to work as intended.

@zicklag @erlend-sh let me know if any concerns with this


zicklag (approved):
Looks good. :+1: I ran into this issue with Dpads earlier, too. I think just adding an alternate controls is a good solution.

PR #992, opened 2024-04-28
fix: Fix machine gun animation and musket

DRuppFv:
The musket simply didn't call `bullet_spawn_offset.y`, no big deal.

The machine gun animation would get frozen when dropped. Had to add an "else" for when the item is not being held, because `items_dropped.get` is called only when in fact the item got dropped, and if the player gets killed while shooting, the animation would still froze, as it wouldn't be IN FACT dropped.

It even made me think about the ammunition reload system of all guns.
```
        if items_dropped.get(entity).is_some() {
            // reload gun
            gun.ammo = *max_ammo;
        }
```
Here, reloading will only take place if the item is dropped. This means that if you die holding a musket with half the ammo, the next player to pick it up will get it with half the ammo. Is that what we really want? Or do we want the ammo to be reloaded?

### DIFF
--- a/src/core/elements/machine_gun.rs
+++ b/src/core/elements/machine_gun.rs
@@ -191,15 +191,14 @@ fn update(
         machine_gun.cooldown.tick(time.delta());
         machine_gun.empty_cooldown.tick(time.delta());
 
+        let sprite = sprites.get_mut(entity).unwrap();
+        // Reset machine gun animation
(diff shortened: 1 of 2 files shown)


MaxCWhitehead:
Good catch. I think we probably want the ammo to be reloaded when player dies and "drops" weapon so it is consistent with deliberately dropping.

PR #993, opened 2024-05-15
feat: Respawn Stamp Boots when a player was killed

### LINKED ISSUE #929: Spike Shoes are still in player's hand after equipping
### Description

After equipping spike shoe they are still in the player's hand and may be dropped, while remaining on player's feet.

### To Reproduce

Pickup spike shoe (Level 3 should have some).

### Expected Behavior

Spike shoe should be consumed once equipped?

### Additional Context

_No response_

### Log Messages

_No response_

Breadp4ck:
This PR closes #929.

I made `WornStompBoots` component that marks the entity that created when the player used the boots. When the player is killed (`PlayerKilled`) and a related entity with `WornStompBoots` exists (`stomp_boots` in `WearingStompBoots`), I respawn the entity with `StompBoots` component and remove marker `WornStompBoots`.

### DIFF
--- a/src/core/elements/stomp_boots.rs
+++ b/src/core/elements/stomp_boots.rs
@@ -202,7 +202,7 @@ fn update_wearer(
                 }
             }
             // We don't need to proceed because the player is dead
-            return;
+            continue;
         }
 
         let kinematic_body = kinematic_bodies.get(entity).unwrap();
--- a/src/core/elements/stomp_boots.rs
+++ b/src/core/elements/stomp_boots.rs
@@ -27,10 +27,15 @@ pub fn session_plugin(session: &mut Session) {
 /// Marker component added to things ( presumably players, but not necessarily! ) that are wearing
 /// stomp boots
 #[derive(Debug, Clone, Copy, Default, HasSchema)]
-pub struct WearingStompBoots;
+pub struct WearingStompBoots {
+    stomp_boots: Entity,
+}
 
 #[derive(Copy, Clone, Debug, HasSchema, Default)]
 pub struct StompBoots;
+
+#[derive(Copy, Clone, Debug, HasSchema, Default)]
+pub struct WornStompBoots;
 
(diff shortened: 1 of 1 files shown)


src/core/elements/stomp_boots.rs
```diff
@@ -167,13 +182,29 @@ fn update(
...
+        if killed_players.get(entity).is_some() {
+            // Respawn the boots worn by the player who was just killed
+            if let Some(spawner) = spawners.get(*stomp_boots) {
+                if worn_stomp_boots.get(*stomp_boots).is_some() {
+                    worn_stomp_boots.remove(*stomp_boots);
+                    hydrated.remove(**spawner);
+                }
+            }
+            // We don't need to proceed because the player is dead
+            return;
```
MaxCWhitehead, line 205: I think this should be `continue` instead of `return`. This would prevent update for a second player who is wearing boots - but we only want to skip this iteration for this killed player.

Otherwise this change looks great, tested it out seems to be working. Thank you!

MaxCWhitehead (changes requested):

src/core/elements/stomp_boots.rs
Breadp4ck: > I think this should be `continue` instead of `return`. This would prevent update for a second player who is wearing boots - but we only want to skip this iteration for this killed player.

I forgot it's not bevy :/

However, after experimenting for a while, I discovered that players can equip two boot items, but only one of them will respawn when the player is killed. I suggest dropping the worn boots when a player attempts to equip another pair.

MaxCWhitehead (approved):

Breadp4ck:
@MaxCWhitehead

I think we should not close the issue right now.

> However, after experimenting for a while, I discovered that players can equip two boot items, but only one of them will respawn when the player is killed. I suggest dropping the worn boots when a player attempts to equip another pair.

PR #996, opened 2024-05-18
feat: upgrade bones and use `iroh` based multiplayer

dignifiedquire:


dignifiedquire:
the clippy failures seem unrelated, let me know if you want me to fix them here

src/ui/network_game.rs
```diff
@@ -456,7 +459,8 @@ pub fn network_game_menu(
                                 .clicked()
                                 {
                                     *status = NetworkGameStatus::Searching;
-                                    online::start_search_for_game(matchmaking_server.clone(), *player_count);
+                                    let server = matchmaking_server.parse().expect("invalid server id");
```
dignifiedquire, line 462: this should really not panic but show an error in the UI

dignifiedquire: but I don't exactly know how, so I'll leave that up to you ;)

MaxCWhitehead: yeah that's fine - we don't have great usage of friendly error display yet to model around.

Cargo.toml
```diff
@@ -152,6 +152,6 @@ ignored = [
 
 # # Uncomment for testing during bones development, if you clone bones adjacent to the jumpy
 # # repo.
-# [patch.'https://github.com/fishfolk/bones']
+[patch.'https://github.com/fishfolk/bones']
```
MaxCWhitehead: Super minor comment but could we remove this diff / keep this line commented? :) Thanks

Otherwise I think this looks good to merge.

MaxCWhitehead:
I setup a matchmaker with persistent node id and changed jumpy's default to this: `lokmcpce2uzezmh2iq3vf23cbu3grzrct66qcpdfrizmcrrlk5uq`

@zicklag if we have any reason want to use existing host can do that as well just throwing something in here for now.

Cargo.toml
MaxCWhitehead: fixed this

MaxCWhitehead (approved):

Issue #1001, opened 2024-06-02
Prevent pause menu from interrupting network of online game


MaxCWhitehead:
### Description

Right now the pause menu will stop running game session / network session runner, this can lead to timeouts / disconnects. Perhaps we should not pause the game when pause menu is open in online play, just let it keep running. 

Or could consider pausing for all players, but this is more complex + also may lead to frustration if game is repeatedly paused by other players. Rounds are fast / cheap so having other players keep going while in menu seems acceptable to me.

Somewhat related to https://github.com/fishfolk/jumpy/issues/994 (pausing may be used to repro this, but not direct / sole cause)

erlend-sh:
Yeah, only pausing the single session seems best. We could still let the other players know that a player has paused, e.g. by showing a ⏸️ sign above their character. That way at least there’s a chance of fair play.

MaxCWhitehead:
> Yeah, only pausing the single session seems best. We could still let the other players know that a player has paused, e.g. by showing a ⏸️ sign above their character. That way at least there’s a chance of fair play.

We could also allow pause for everyone as a custom lobby setting or something down the line maybe if desired.

MaxCWhitehead:
This is fixed, pause in online now keeps game going in background + no longer disconnects network. On full disconnects players are now also notified / given button to leave game.

PR #1003, opened 2024-06-09
fix: Prevent pausing in online from timing out network, allow game to continue running

MaxCWhitehead:
Previously pausing in online would pause net loop and cause timeout. Now game continues to simulate in background, and inputs are disabled while pause menu open in online.

Dependent on https://github.com/fishfolk/bones/pull/411, CI will fail until updated.

Fixes https://github.com/fishfolk/jumpy/issues/1001

### DIFF
--- a/src/ui/pause_menu.rs
+++ b/src/ui/pause_menu.rs
@@ -32,15 +32,7 @@ fn pause_menu_system(
     world: &World,
     assets: Res<AssetServer>,
     mut pause_menu: ResMutInit<PauseMenu>,
-
-    #[cfg(not(target_arch = "wasm32"))] socket: Option<Res<NetworkMatchSocket>>,
 ) {
-    // TODO allow pause menu in online game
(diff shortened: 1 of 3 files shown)


src/ui/pause_menu.rs
```diff
@@ -275,3 +271,16 @@ fn main_pause_menu(
...
+
+/// Helper for pausing session depending on if online or offline match.
+/// `remain_inactive` is used if other systems like scoring have set session inactive and wish to
+/// prevent that change.
+fn pause_session(paused: bool, is_online: bool, session: &mut Session, remain_inactive: bool) {
+    if is_online {
+        // Online session remains active so that it will not timeout with other players.
+        // Instead we disable the input so it is not captured by game while paused.
+        session.runner.disable_local_input(paused);
+    } else if !paused && !remain_inactive {
```
MaxCWhitehead, line 283: whoops looks like a bug here, need to make sure branch for un-pausing local sessions is present.

PR #1031, opened 2024-10-01
feat: quick debug env vars

nelson137:
Add support for some environment variables, only available in debug builds, that allow for quickly loading into a single-player game.

## Examples

To launch straight into a single-player match with Fishy, the default player, on Keyboard1:

```shell
TEST_MAP='Dev Level 1' cargo run
```

You can also specify another player:

```shell
TEST_MAP='Dev Level 1' TEST_PLAYER='Sharky' cargo run
```

And even provide a hat:

```shell
TEST_MAP='Dev Level 1' TEST_PLAYER='Sharky' TEST_HAT='Cowboy Hat' cargo run
```

Or, auto-select a player, launching directly to the map select menu (again, optionally with a hat):

```shell
TEST_PLAYER='Orcy' TEST_HAT='Unicorn Hat' cargo run
```

All of these examples set Keyboard1 as the control source for the player. You can provide a different source like so:

```shell
TEST_CONTROLLER='Gamepad' TEST_MAP='Dev Level 1' cargo run
```

Any combination of one or more of these four environment variables will skip straight to either the map select menu or directly into a match if a map is provided. Fishy with no hat on keyboard1 is the default if only a map is given.


MaxCWhitehead (approved):
This is great, thanks!

PR #1037, opened 2024-10-21
fix: require hat select before ready up

### LINKED ISSUE #963: Player Select: Last player readying up cannot choose hat in online play (match starts immediately)
### Description

In online play, it seems hat is selected after readying up, if the last player hits ready, the lobby transitions to map select before player has opportunity to select hat. 

In local play, this is not an issue.

### To Reproduce

_No response_

### Expected Behavior

Should move hat select before ready up?

### Additional Context

_No response_

### Log Messages

_No response_

nelson137:
Fixes #963

## Changes

- [6781542](https://github.com/fishfolk/jumpy/pull/1037/commits/6781542946af7aafdaf0fe69ce0cf7cdd63b0781) `PlayerSlot` is now (sort of) a state machine, starting out `Empty` then progressing through `SelectingPlayer` and `SelectingHat` until becoming `Ready`
    - Slot for a network player only uses the `SelectingHat` and `Ready` states for simplicity
    - Angled brackets `<` and `>` show around the thing being selected (player/hat), brackets now don't show when ready
- [a2393c4](https://github.com/fishfolk/jumpy/pull/1037/commits/a2393c462d54a45c3b66d92d1b02559b435504aa) fix bug where label has incorrect text for 1 frame when pressing Menu Back to remove player from slot
- [f636d33](https://github.com/fishfolk/jumpy/pull/1037/commits/f636d3308832777d6714bceaba18c7e9e169da15) Upgrade bones to latest
- Set gitattributes for `*.ftl` files so that they are treated as text for diffs and not binary files
- Ignore advisory `RUSTSEC-2024-0370` proc-macro-error is unmaintained (see [related bones issue](https://github.com/fishfolk/bones/issues/479) and [the advisory](https://rustsec.org/advisories/RUSTSEC-2024-0370))

This also fixes a bug I was seeing on `main` where another player in a LAN lobby pressing A/D would behave as if everyone pressed A/D, e.g. pressing D to select the next skin would change to the next skin in both slots.

**Rationale for the rough state machine:**

Previously there were 3 states for a player slot: Empty, SelectingPlayer, and Ready + SelectingHat. This state was maintained by two booleans, `active` and `confirmed`, which tell you whether the slot is not empty and whether the player has readied-up respectively. Together they represent the 3 states as: (`active/confirmed`) `false/false`, `true/false` and `true/true`; with `false/true` being an invalid state.

Now that there are four states to the ready-up process, one way of implementing this is to add a third boolean, but this introduces 3 more invalid states. Let's say the boolean is called `selectedPlayer`, these are the possible boolean combinations and the states they represent (`active/selectedPlayer/confirmed`):

```
false/false/false Empty
false/false/true <invalid>
false/true/false <invalid>
false/true/true <invalid>
true/false/false SelectingPlayer
true/false/true <invalid>
true/true/false SelectingHat
true/true/true Ready
```

The implementation I opted for is a simple state machine. While it added a lot more code to some areas of this module I found it made reasoning about some sections much easier. It also made explicit certain edge cases in the `handle_match_setup_messages` system where e.g the client could theoretically receive a confirmation message from a peer when it had not yet received any player selection messages. A warning log is now generated in the unlikely case this occurs.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -21,5 +21,7 @@ web/* linguist-vendored
 assets/** binary
 old_assets/** binary
 
+*.ftl diff
+
 # If you have other directories with mostly binary files, add them like this:
 # other_binary_directory/** binary
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -76,6 +76,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "acto"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "31c372578ce4215ccf94ec3f3585fbb6a902e47d07b064ff8a55d850ffb5025e"
+dependencies = [
+ "parking_lot",
+ "pin-project-lite",
+ "rustc_version",
+ "smol_str 0.1.24",
+ "tokio",
+ "tracing",
+]
+
+[[package]]
 name = "addr2line"
 version = "0.24.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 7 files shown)


MaxCWhitehead (approved):
Looks good to me, thanks! I agree that using enum seems more clear than stacking more bools 👍

MaxCWhitehead:
> This also fixes a bug I was seeing on main where another player in a LAN lobby pressing A/D would behave as if everyone pressed A/D, e.g. pressing D to select the next skin would change to the next skin in both slots.

Nice catch on that!

PR #1040, opened 2024-11-02
fix(player_select): Fix Online game able to start without all playes ready 

### LINKED ISSUE #1036: Online lobby goes to map select once local player is ready, before all players are ready
### Description

When starting an online match, one your player is readied up (but not remote player), it seems it goes to map select before all players are ready.

I am unsure if this is a bug I introduced with https://github.com/fishfolk/jumpy/pull/1032 or not. It is also possible it is an issue with bones matchmaker changes, but probably an issue in Jumpy, should inspect that first.

### To Reproduce

1. Start online match with two players
2. Press ready on either one (but only client)
3. Observe this client is now on map select, and second is still on player select.

(From this point selecting a map / proceeding may cause a crash due to network asset cid mismatch).

If these steps don't repro, it's possible it needs to be done more than once. (Though if we find this is the case, that may indicate this is impacted by some bones matchmaker state and not just jumpy usage of it).

### Expected Behavior

Should not advance to map select until all players are ready.

### Additional Context

_No response_

### Log Messages

_No response_

MaxCWhitehead:
Fixes #1036 and couple other things along the way

Not entirely sure what caused this - but I think I introduced it when updating for new matchmaker interface in bones.

We would allow online matches to continue if >1 players are ready and no unconfirmed players. The issue is that in online remote players who had not made a selection were in `Empty` state, and not tracked as unconfirmed. Now for any local + required players, I initialize them to new PlayerSlot type `SelectingLocalControlSource`. `Empty` may be used for online slots that are unused (the last 2 slots in a 2 player game) in online, or for lan, optional slots that may be joined by additional players/AI, but are not required. And remote required players are set to `SelectingPlayer` with default fish + no hat - as next message expected from them is change in player, and they don't need to select control source, we know they are remote.

This seems to have fixed the match starting without all players ready, as all required players are initialized to something other than `Empty`.

#### Remembering Hats:

I noticed that if you select a hat and then deselect it going back to fish select - the other online players would still see this hat, but local player resets to None. If player selected hat, went back to fish select, then readied up with no hat, the remote players would see hat, but not the local player. This could cause desync. Now When you go back to fish select, your hat is remembered. 

#### Always display default fish:
We were no longer displaying default fish in online when player had not yet selected a control source/fish. Now regardless of if have actually selected a fish, the default fish + no hat is shown (If the slot is a required player that we are sure will select one eventually / must ready up). In local for unjoined slots, no default fish is shown until joined or AI is set.

#### Show when remote players ready
In online when a remote player is ready, this is now displayed.

@nelson137 feel free to review (if you want) as I know you were just in here. Player select sure is complicated lol, but feeling good about the changes I think we're making it more manageable 😅

### DIFF
--- a/src/ui/main_menu/player_select.rs
+++ b/src/ui/main_menu/player_select.rs
@@ -67,6 +67,7 @@ pub enum PlayerSlot {
     SelectingPlayer {
         control_source: PlayerSlotControlSource,
         current_player: Handle<PlayerMeta>,
+        current_hat: Option<Handle<HatMeta>>,
     },
     SelectingHat {
         control_source: PlayerSlotControlSource,
@@ -152,7 +153,8 @@ impl PlayerSlot {
 
     pub fn selected_hat(&self) -> Option<Handle<HatMeta>> {
         match self {
-            Self::Empty | Self::SelectingLocalControlSource | Self::SelectingPlayer { .. } => None,
+            Self::Empty | Self::SelectingLocalControlSource => None,
+            Self::SelectingPlayer { current_hat, .. } => *current_hat,
             Self::SelectingHat { current_hat, .. } => *current_hat,
             Self::Ready { selected_hat, .. } => *selected_hat,
         }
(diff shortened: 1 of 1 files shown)


nelson137:
Nothing to add here, I didn't test it but these look like solid changes!

PR #1041, opened 2024-11-07
fix(sprite): reposition rope tiles in atlas for padding

### LINKED ISSUE #1014: Rendering artifacts on edge of rope netting sprite
### Description

image

I have been seeing some thin black lines to the left of the rope netting. Might be related to texture sampling settings, or maybe with atlas?

EDIT:
It seems bevy defaults to nearest sampling which seems right. It seems to flicker in and out as camera moves, wonder if we have mip maps and if that would help.

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Additional Context

_No response_

### Log Messages

_No response_

nelson137:
Fixes #1014 (for now).

It sounds like a proper fix entails significant work; but to fix the immediate issue we can move the rope tiles away from any others so that those adjacent to them are blank.

## Before

before-asset
before-game

## After

after-asset
after-game


MaxCWhitehead:
Yeah a rendering fix did not seem very clear, I think this is good, I had not tried this but good to know can be fixed in authoring of the tilesets.

Haven't had a chance to boot up game and double check - I see level 3 is updated. Do you happen to know if any other levels use these rope tiles or just 3?

If no other levels will go ahead and merge. Thanks!

nelson137:
Good catch, levels 7 & 8 also have ropes that were broken. Those are now fixed:

## Level 7

| Before | After |
| --- | --- |
| level-7-before | level-7-after |

## Level 8

| Before | After |
| --- | --- |
| level-8-before | level-8-after |

nelson137:
Well I found a few more, the railing and flag in this atlas have the same issue.

MaxCWhitehead:
> Well I found a few more, the railing and flag in this atlas have the same issue.
> 
> EDIT: and the crow's nest

Ah gotcha, those were much less noticeable to me. Good to know. Should I go ahead and merge this (we can open a new issue for those)? If want to update PR for anything I can leave it open.

MaxCWhitehead (approved):

nelson137:
Yeah I think this is good to merge as-is. I can do the others in a follow-up.

PR #1042, opened 2024-11-08
fix(ui): scroll to focused button in map select

### LINKED ISSUE #976: Navigating menus (map select) with gamepad/keyboard does not scroll when moving selection outside visible item
### Description

On a menu with a list of items such as map select, depending on scaling there may be a scroll bar. If moving focused selection with keyboard or gamepad, when moved to bottom outside visible area, it does not scroll view down. Scrolling with mouse is required to see what is selected.

### To Reproduce

1. Go into non-fullscreen mode
2. Start game and get to map select
3. press down until focus is outside of viewable items, note that scroll did not occur

### Expected Behavior

Scrolling lists should automatically scroll to keep focus in view

### Additional Context

_No response_

### Log Messages

_No response_

nelson137:
Fixes #976.

### DIFF
--- a/src/ui/map_select.rs
+++ b/src/ui/map_select.rs
@@ -81,6 +81,10 @@ pub fn map_select_menu(
                                     .show(ui)
                                     .focus_by_default(ui);
 
+                                    if all_maps_button.has_focus() {
+                                        all_maps_button.scroll_to_me(None);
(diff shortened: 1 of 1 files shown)


MaxCWhitehead:
Sweet, thanks!

PR #1044, opened 2024-11-17
fix: sprite artifacts 2

nelson137:
Follow-up to #1041.

## Changes

- Fix the rest of the sprite artifacts in `ship_decorations`

## Things Worth Mentioning

- Changed levels 3 & 8 to use YAML flow style so that I could find tiles by position (e.g. by searching for `pos: [13, 10]`:

    Before: 
    ```yaml
    tiles:
    - pos:
      - 1
      - 2
      idx: 3
      collision: Solid
    ```
    After:
    ```yaml
    tiles:
    - { pos: [1, 2], idx: 3, collision: Solid }
    ```

- Add script to get the layer names in any level files that use specific tile indexes of a tilemap:

    ```yaml
    # Find all crow's nests (indexes 30-32) from ship_decorations
    $ ./scripts/find-atlas-tiles.sh ship_decorations.atlas.yaml '[30,31,32]'

    ./assets/map/levels/level_3.map.yaml
    - foreground

    ./assets/map/levels/level_8.map.yaml
    - foreground

    # This tells us levels 3 and 8 have crow's nests in their "foreground" layer
    ```

- Fix git attributes for text files in `assets/` & `old_assets/` which shouldn't be treated as binary (yaml, txt, ftl)


MaxCWhitehead:
Sorry fixing this up is so painful - need to get the level editor back in. Thanks!

PR #1051, opened 2025-06-29
chore(deps): bump rapier2d to 0.19.0

marcin-serwin:
The forked version wasn't working for me when building in a sandbox, so I've bumped it to the upstream one that contained the fix.


zicklag:
Looks good, thanks!

